PR #2, opened 2021-11-13
Changes name to @openhab/automation, adds webpack to generate a single file

digitaldan:

### DIFF
--- a/log.js
+++ b/log.js
@@ -3,7 +3,7 @@
  * This namespace provides loggers to log messages to the Openhab Log.
  * 
  * @example <caption>Basic logging</caption>
- * let log = require('ohj').log('my_logger');
+ * let log = require('@oh').log('my_logger');
  * log.info("Hello World!")
  * 
  * @namespace log
@@ -209,11 +209,11 @@ let getAppenderForLevel = function(){
     if(appenderForLevel === null) {
         appenderForLevel = () => null;
 
-        try {
-            appenderForLevel = require('log_appenders').forLevel;
-        } catch(e) {
-            new Logger("log", () => null).debug("No appenders found for log", e);
-        }
+        // try {
+        //     appenderForLevel = require('log_appenders').forLevel;
+        // } catch(e) {
+        //     new Logger("log", () => null).debug("No appenders found for log", e);
+        // }
     }
 
     return appenderForLevel;
--- a/log.js
+++ b/log.js
@@ -3,7 +3,7 @@
  * This namespace provides loggers to log messages to the Openhab Log.
  * 
  * @example <caption>Basic logging</caption>
- * let log = require('@oh').log('my_logger');
+ * let log = require('@openhab/automation').log('my_logger');
  * log.info("Hello World!")
  * 
  * @namespace log
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 node_modules
 out
+dist
--- a/README.md
+++ b/README.md
@@ -1,9 +1,9 @@
-[![Build Status](https://travis-ci.org/jpg0/ohj.svg?branch=master)](https://travis-ci.org/jpg0/ohj)
+[![Build Status](https://travis-ci.org/openhab/openhab-js.svg?branch=master)](https://travis-ci.org/openhab/openhab-js)
 # Openhab Javascript Library
 
 This library aims to be a fairly high-level ES6 library to support automation in Openhab.
 
-[API Documentation](https://jpg0.github.io/ohj/ohj/0.1.1/)
+[API Documentation](https://openhab.github.io/openhab-js/oh/0.1.1/)
 
 ## Requirements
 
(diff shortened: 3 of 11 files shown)


log.js
```diff
@@ -209,11 +209,11 @@ let getAppenderForLevel = function(){
     if(appenderForLevel === null) {
         appenderForLevel = () => null;
 
-        try {
-            appenderForLevel = require('log_appenders').forLevel;
-        } catch(e) {
-            new Logger("log", () => null).debug("No appenders found for log", e);
-        }
+        // try {
```
digitaldan, line 212: Hi @jpg0 I commented this out as i'm not sure where log_appenders is coming from, is this needed?

digitaldan:
I have another PR i'll hold off on submitting until this gets merged.

digitaldan:
Hmm, so @jpg0  i think its going to be tricky to publish a "scoped" module like this to NPM.  Not sure if you have an idea in mind, but i think you can only publish packages under this scope, for example `@oh/js` or `@oh/items` but not the actual root scope/package on its own like just `@oh`.    I guess one option is to publish all the individual  members (items, things, etc...) as packages with dependencies on each other, but this seems messy and complicated.  Another option is to publish under the scope, so `@oh/automation`, `@openhab/automation` or something similar , or we could publish as some plain unscoped NPM name. 

https://docs.npmjs.com/creating-and-publishing-scoped-public-packages

log.js
jpg0: The original use-case here was to allow custom log appenders for JS code. Specifically, in my case, to log errors to Telegram. I didn't want to require boilerplate in each JS script, so I put it into the log library, and had it "optimistically" look up a custom module to search for extra appenders. In my installation I provided an appender to log to Telegram.

I don't believe that anyone else has used this functionality though, so I would say that it's fine to remove.

jpg0:
Re: scoped modules - I guess the challenge here is that we want (a) no collisions with existing modules and (b) concise scripts. Saying that, if we automatically include the imports for UI scripts, is the second point required any longer? Maybe just go for `openhab-js` in this case?

There is also the option to remap `@oh` to something longer, but I'd avoid this as it's likely to break other things.

digitaldan:
>if we automatically include the imports for UI scripts, is the second point required any longer? 

I think longer is ok here as well.  I wonder if we should create an official @openhab scope on NPM and publish under this? I like the clarity of `@openhab/automation` , although it's a tad bit long.   Unfortunately there is already a `openhab` package on NPM that prevents us from claiming the `@openhab` organization/scope.  I have reached out to the author of that module about giving us the name as his library has not been updated in several years.

digitaldan:
So the developer of the `openhab` npm package has graciously agreed to release this to us, i think since this is an "official" openhab NPM, we should probably publish under that.  Thoughts?

jpg0:
Yes, I agree.

digitaldan:
any thoughts on package name?  `@openhab/rules` , `@openhab/automation`, or something different?  Technically we could publish as just `openhab` if we wanted to.

digitaldan:
Also i noticed that https://github.com/statelyai/xstate manages to publish to both a default package to `xstate` and also use scopes for sub packages like `@xstate/fsm` , so it is possible to mix.

jpg0:
One thing to be aware of is that often scripts are run fairly regularly, and in this case there is little caching of code afaik. I do think that it would be worth ensuring that not all of the library needs to be parsed to run any script. I guess this can be done either my splitting it into multiple packages, or by importing them only when used.

As for the actual name, I don't have a strong opinion, other than in trying to make it as future-facing as possible. One thing that I would point out is that I actually used JS for more than rules in my system - I also used it for the bulk of the item and thing definitions, so I guess more than even 'automation'.

digitaldan:
> I also used it for the bulk of the item and thing definitions, so I guess more than even 'automation'.

How are you doing that?  Is it still through the JSBinding?

FYI i just received ownership of the openhab npm package.

jpg0:
Yes, through the binding. The script to insert things/items/links is https://github.com/jpg0/oh-config/blob/master/automation/jsr223/javascript/personal/_25_things_and_items.sl25.js - this repo has my old scripts in it; there is quite a lot there if you're interested.

digitaldan:
Got it, i will still argue `automation` is a relevant term then since this is happening in an automation addon

>One thing to be aware of is that often scripts are run fairly regularly, and in this case there is little caching of code afaik

I was looking at how to share variables across contexts (to support references to timers and such between reloading of a script) and came across this,  quickly reading this leads me to believe that if we reuse an engine we might get caching out of the box?  I'm guessing they do some sort of hash to see if the loaded code has changed since the last execution ? 

https://www.graalvm.org/reference-manual/embed-languages/#code-caching-across-multiple-contexts

jpg0:
`automation` is fine by me.

We may get caching if we reuse an engine - I have also not looked into it. I opted for engine-per-script originally as that was how it all worked with the existing `javax.script` stuff.

digitaldan:
Oof, so we have another problem, while we have the package name "openhab" on NPM now, we can not publish to the same scope as there is a user called "openhab" already registered.  I have submitted a ticket to npmjs support about claiming the username so we can own the scope. 

I may change this to just `openhab` for the time being as i'm not very confident they will transfer a user name to us.

PR #4, opened 2021-11-16
This adds new trigger channels, as well as some small OH3 compatabili…

digitaldan:
…ty updates

### DIFF
--- a/fluent/fluent.js
+++ b/fluent/fluent.js
@@ -118,6 +118,15 @@ const fluentExports = {
     timeOfDay: s => new triggers.ItemTriggerConfig('vTimeOfDay').changed().to(s),
     
     /**
+     * Specifies a channel event as a source for the rule to fire.
+     * 
+     * @memberof fluent
+     * @param {String} channelName the name of the channel
+     * @returns {ItemTriggerConfig} the trigger config
+     */
+    channel: s => new triggers.ChannelTriggerConfig(s),
+
+    /**
      * Specifies a cron schedule for the rule to fire.
      * 
      * @memberof fluent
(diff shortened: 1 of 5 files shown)


fluent/fluent.js
```diff
@@ -207,7 +216,16 @@ const fluentExports = {
...
-    
+
+    /**
+     * Specifies an group member as the source of changes to trigger a rule.
+     * 
+     * @memberof fluent
+     * @param {String} groupName the name of the group
+     * @returns {ItemTriggerConfig} the trigger config
+     */
+    memberOf: s => new triggers.ItemTriggerConfig(s, true),
```
digitaldan, line 227: FYI memberOf is how this is used in our classic DLS rules engine as welll

fluent/trigger-conf.js
```diff
@@ -130,7 +171,115 @@ class ItemTriggerConfig {
...
+    constructor() {
+        this._toOHTriggers = () => [triggers.SystemStartlevelTrigger(this.level)]
+        this.describe = (compact) => compact ? `system:${this.level}` : `system level "${this.level}"`
+    }
+    _complete() {
+        return typeof (this.level) !== 'undefined';
+    }
+
+    rulesLoaded() {
+        this.level = 40;
```
digitaldan, line 249: I should probably just call `return this.startLevel(40)` for this and the other methods

digitaldan:
@jpg0 if you could review that would be awesome!

jpg0 (approved):

jpg0:
FYI you may want to remove the `utils.typeWithFallback` stuff - this was to explicitly support both OH 2 & 3 concurrently as things moved packages. It probably isn't necessary any longer as we are only targeting 3.

digitaldan:
>FYI you may want to remove the utils.typeWithFallback

Agreed,  thanks for reviewing!

PR #6, opened 2021-11-27
Rule builder API

digitaldan:
This converts the fluent package into the  "Rule Builder"


digitaldan:
@jpg0 Here is the proposed Rule API , i also have started working on updating the documentation (which is a very much work in progress).  Would be great if you could take a look and give feedback on the builder api.  I also have a few questions which i will ask as comments on this PR for you.  Thanks!

rules/operation-builder.js
```diff
@@ -0,0 +1,481 @@
...
+        this._then(this.fn, this.group, name, description);
+    }
+
+    /**
+     * Specify the rule group for this rule
+     * 
+     * @param {string} group the group this rule belongs to.
+     * @returns {OperationBuilder} this
+     */
+    inGroup(group) {
```
digitaldan, line 45: @jpg0 I believe the rule group functionality is something not native to openHAB, and that this library is artificially constructing this feature, correct?  I'm wondering if this should be removed (and the rule toggle features) for simplicity sake?  While a nice feature, it seems like it would be out of scope for what is the official library/API which shouldn't add such new functionality on its own.

rules/operation-builder.js
```diff
@@ -0,0 +1,481 @@
...
+        this.to_item = item_name;
+        return this;
+    }
+
+    /**
+     * Appends another operation to execute when the rule fires
+     * 
+     * @returns {OperationBuilder.CopyStateOperation} this
+     */
+    and() {
```
digitaldan, line 228: I have not really tried this yet, so not sure if it will work

digitaldan:
I'm going to merge this as its a isolated change to the rest of the API, please feel free to comment on it i'm sure it will change

PR #25, opened 2021-12-07
Pull the MetadataRegistry from OSGI

rkoshak:
The methods we need to interact with the MetadataRegistry are not static. We need to pull an instance of the registry from OSGI to get, set and update metadata.

Signed off by Richard Koshak <rlkoshak@gmail.com>


rkoshak:
Grrr, that's the last time I try to do it through the Web UI. The DCO should be passing now.

digitaldan:
LGTM, thanks!

PR #27, opened 2021-12-12
adds all persistence methods along with the ability to use a persistence ID to Items

digitaldan:


digitaldan:
Screen Shot 2021-12-12 at 11 19 40 AM

digitaldan:
@jpg0 the other thing i will mention is that while i have these wrappers as part of a class that OHItem extends, thats probably not necessary and might be best to merge them into the OHItem class itself ? 

@rkoshak let me know if these look usable, all the methods for persistence are directly on the OH Item in this PR, so it would look like:
```javascript
const item = items.getItem("Lights");
//default persistance
item.historicState​();

//influxdb  persistance
item.historicState​("influxdb");
```

jpg0:
Personally, I am a little fearful of merging all these methods onto `item` directly.

My experience with APIs is that you need to be fairly careful on what you expose because you are unlikely to be able to fix it later (once it's in use).

In this case, the API is structurally diverging from the core openHAB API quite a bit. The dangers are that (a) a conflict may arise in methods in the two areas (for example if there was a number of similar methods introduced on each object in core, that were awkward to merge together here), or (b) the core APIs are restructured - for example it became "itemAndChannelHistory" (far-fetched example, I know), or was split across multiple classes - making it no longer 1:1 with the items.

I don't regard the possibility of this happening is high, so maybe it's just my paranoia, but I have certainly seen problems arise in the past where API layers have structurally diverged too far.

The alternative would be to put all historic methods on a child object like '.history' or whatever. Either way I'd support whichever way you want to go though.

(Oh, and slightly OT - I had always meant to rename `OHItem` to simply `Item`, but never got around to it. It's not a big deal as `OHItem` does not appear in the API, but does show up as `OHItem` in stack traces.)

rkoshak:
Given how the APIs are built I like having them be a part of something that gets documented, whether these are part of OHItem or completely separate.

Looking through the methods I don't have any major concerns beyond the fact that we are adding a whole bunch of new methods to the OHItem class which might become unwieldy. Functionally though it looks like everything is there (including some I wasn't aware were implemented at all like deviation and evolution).

In regard t your concern @jpg0, I share that concern. However, the Persistence API methods have not changed except for addition to it since OH 1.6 so the likelihood of the differences coming to bite us in the end is probably pretty low. But it's not zero. Also, any change to this will break all rules, not just JSScripting so it would be a pretty big deal and there would need to be a bunch of OH wide coordination to pull it off.

Personally I'm not the biggest fan of spooky magic when it comes to APIs though. When we make things like sendCommand and lastUpdate appear to be a part of the Item when in fact they are separate, it sets up a false understanding and expectations which can make debugging and understanding one's own rules difficult. I've seen this a number of times on the forum and experienced it some myself (though that was more getting my head around Jython and Nashorn JS). I don't feel strongly about this either way because there are definite advantages to making these appear to be a part of an Item too. I just wanted to throw it out there for consideration.

digitaldan:
So I think i'm leaning towards creating a ItemHistory class that would be accessed from `someitem.history.average(...)` .  We can start there and see how this looks, it will be easy to move on the item if we need to.   Also i like renaming OHItem to Item.

digitaldan:
So docs look like this which seems pretty good

image

digitaldan:
I'm going to merge this, but if anyone has other thoughts , please let me know, its easy enough to change!

PR #39, opened 2021-12-17
Add math and count operations on groups

florian-h05:
@digitaldan 
This PR includes most parts of the [groupUtils](https://github.com/rkoshak/openhab-rules-tools/tree/main/group_utils) from @rkoshak 's [openhab-rules-tools](https://github.com/rkoshak/openhab-rules-tools) in the helper library.
NOTE: I am the author of them, so copyright is no problem.

It includes:

- Get an array of only the names of the group members. Useful for iterating on group members.
- Get a concatenated string of the members' labels. Useful for UI and notifications.
- Get the member item with the minimum or maximum state.
- Get the avg or sum of members' states. (A filter for !`isUninitialized` and Item types Number, Dimmer & Rollershutter is applied).
- Cound how many members match a given comparison function.

I also fixed some typos in the JSDoc comments.

### DIFF
--- a/items/managed.js
+++ b/items/managed.js
@@ -26,7 +26,7 @@ const DYNAMIC_ITEM_TAG = "_DYNAMIC_";
  * 
  * @memberOf items
  */
- class Item {
+class Item {
     /**
      * Create an Item, wrapping a native Java openHAB Item. Don't use this constructor, instead call {@link getItem}.
      * @param {HostItem} rawItem Java Item from Host
@@ -55,7 +55,7 @@ const DYNAMIC_ITEM_TAG = "_DYNAMIC_";
 
     /**
      * The name of the item.
-     * @return {String} the name
+     * @returns {String} the name
      */
     get name() {
         return this.rawItem.getName();
@@ -63,7 +63,7 @@ const DYNAMIC_ITEM_TAG = "_DYNAMIC_";
 
     /**
      * The label attached to the item
-     * @return {String} the label
+     * @returns {String} the label
      */
     get label() {
         return this.rawItem.getLabel();
@@ -71,7 +71,7 @@ const DYNAMIC_ITEM_TAG = "_DYNAMIC_";
 
     /**
      * The state of the item, as a string.
-     * @return {String} the item's state
+     * @returns {String} the item's state
      */
     get state() {
         return this.rawState.toString();
@@ -79,7 +79,7 @@ const DYNAMIC_ITEM_TAG = "_DYNAMIC_";
 
     /**
      * The raw state of the item, as a java object.
-     * @return {HostState} the item's state
+     * @returns {HostState} the item's state
      */
     get rawState() {
         return this.rawItem.state;
@@ -87,12 +87,10 @@ const DYNAMIC_ITEM_TAG = "_DYNAMIC_";
 
     /**
      * Members / children / direct descendents of the current group item (as returned by 'getMembers()'). Must be a group item.
-     * 
+     * @returns {Item[]} member items
      * @example
      * const group = items.getItem('test_group').members;
      * for (let i = 0; i < group.length; i++) console.log(group[i].label);
-     * 
-     * @returns {Item[]} member items
      */
     get members() {
         return utils.javaSetToJsArray(this.rawItem.getMembers()).map(raw => new Item(raw));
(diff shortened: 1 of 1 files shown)


digitaldan:
Thanks for the PR!  This adds a lot (14?) of connivance methods for filtering/transforming groups purely in Javascript, but not matching any server side features (like if the API offered these and we were wrapping those calls) .  I wondering where we draw the line on "syntactic sugar" vs core functionality ?  Since this is all on the `item` namespace, it's going to create a lot of noise in the documentation and autocomplete features in the UI  ( this is of course my opinion).

I would be interested in others to hear what we include as "core" in this library and what should remain user supplied scripts.   I would also be open to alternatives as well , like having connivence namespace for helpers much like we did with `item.history`  .

Again looking for a discussion on these types of matters since we do not have a lot of precedence here yet :-) 

@jpg0 @rkoshak  @Confectrician ?

rkoshak:
I haven't had a chance to compare what's already in the library and what's in this PR. I think there might be a bit if overlap between the two.

I too am concerned about too much being added to items. There is still much there already.

Since theses only apply to members if a group, would there be a way to append the methods to the Array returned by members?

florian-h05:
digitaldan:
> I wondering where we draw the line on "syntactic sugar" vs core functionality ?
>
> I would be interested in others to hear what we include as "core" in this library and what should  remain user supplied scripts.

Yeah, I understand your point.
I think that the library should include stuff most users would find helpful and not only wrap the core API, but it clearly is no it-can-do-everything library.

> I would also be open to alternatives as well , like having connivence  namespace for helpers much like we did with item.history .

I was also thinking about the way of adding those when I wrote them, I had in mind of putting them in something like `items.grouputils` or `items.group` like with `item.history`.  

rkoshak:
> Since theses only apply to members if a group, would there be a way to append the methods to the Array returned by members?

I think that would be the most intuitive way.

jpg0:
My view (apologies in advance) is that I don't think that these kind of things belong here. However many methods you choose to add to conveniently manipulate a list of group members, I can come up with more requirements of other ways I may want it to be done that are not covered (what if I want to exclude some members?). This is why JS has the fundamental concepts only (looping & transforming), which allow users to do it however they want. My belief is that this approach, to build the core, simple concepts than can be combined and reused to manipulate the data any way the user likes, is the right way to go. I don't think think that there are many examples of successful APIs where this level of "conveniences" are included.

Ultimately I think this is another example of providing a pure development environment for Javascript, versus trying to make things as easy as possible for non-developers by trying to anticipate what they are going to need and doing it for them already (the same reason I'm against built-in globals tbh!). 

When I look at what the functions provided offer, they seem to fall into a few categories: mapping the members based on a property, joining these things into a string, or reducing across the items based on some logic.

For mapping, why can't we just use `item.members.map(i => i.name)` (it's barely longer than `items.members.getNames()`. If there are some properties than cannot be mapped for some reason, let's ensure that they can.

For joining into a string, the pure JS is also pretty compact: `item.members.map(i => i.name).join(' ')`. It is also clearly much more flexible (e.g. choice of delimiter).

Reduction is a little more interesting; while I still think it shouldn't be included, it points at a potential lack of supported functionality on items (or their states).

For example, to get the max value for members. I don't agree that we should explicitly include the reduction across items to calculate the maximum, however I do recognise that the function to either extract the state as a number value (or comparable value) or to compare (or order) states is something that we _should_ provide. This function could then be used when looping (or reducing across) the members of a group to find the max (but could also by used much more wider in other use-cases too).

florian-h05:
> My view (apologies in advance) is that I don't think that these kind of things belong here. However many methods you choose to add to conveniently manipulate a list of group members, I can come up with more requirements of other ways I may want it to be done that are not covered (what if I want to exclude some members?).

> Ultimately I think this is another example of providing a pure development environment for Javascript, versus trying to make things as easy as possible for non-developers by trying to anticipate what they are going to need and doing it for them already (the same reason I'm against built-in globals tbh!).

You definitively have a point with purity.
If we do not include those functions, the stuff they provide will be something to write about in a But how do I? section of the docs.

digitaldan:
>If we do not include those functions, the stuff they provide will be something to write about in a But how do I? section of the docs.

I think this is the way to go, and to be honest, might be the most important part of the entire readme.  Most users are going to jump to the examples first, and then go back to the rest of the doc later.  I think we  should add a "Examples" section with subsections for Items, Groups, Actions, Rules, Time, Timers,  etc......

florian-h05:
Okay, then I am going to close this PR.

I think that I will start working on parts of the „Examples“ section and add examples for what the functions (from this PR) do. For those who are interested, I will put those functions in my personal tools library where I collect some very special stuff at [florianh-openhab-tools](https://www.npmjs.com/package/florianh-openhab-tools).

PR #42, opened 2021-12-21
Added key to the getMetadataValue to be able to read namespace config…

marcinczeczko:
I was experimenting with the JS Scripting and found out that the Item has missing the ability to read the metadata configuration value - might be useful in rules such as Times of the Day.

This PR extends the `Item.GetMetadataValue` method to accept `key` of the configuration value within the given `namespace` - if not specified the metadata namespace value will be returned as before.

### DIFF
--- a/items/managed.js
+++ b/items/managed.js
@@ -120,10 +120,11 @@ class Item {
     /**
      * Gets metadata values for this item.
      * @param {String} namespace The namespace for the metadata to retreive
+     * @param {String} key The configuration key of the given namespace to retrieve
      * @returns {String} the metadata associated with this item and namespace
      */
-    getMetadataValue(namespace) {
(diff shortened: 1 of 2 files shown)


digitaldan:
I'm not sure overloading here is the right way.  I wonder if we should enhance metadata.js to support configuration options, then expose the class onto the item, so `item.metadata.getValue(namespace)` and ` item.metadata.getConfiguration(namespace,key)` ?  WDYT?

digitaldan:
Take a look at PR #43 and let me know if this works.  I also like moving all these functions off the item and into its own namespace

marcinczeczko:
@digitaldan sounds good to me. I like namespacing the stuff. Anyway, so looks like I can close my PR. Anyway, I do have one comment to your PR, so will post it over there. Thanks

PR #43, opened 2021-12-22
Moves metadata to its own class accessed through Items

digitaldan:
also adds getConfigurationValue

### DIFF
--- a/metadata/metadata.js
+++ b/metadata/metadata.js
@@ -7,70 +7,113 @@
  */
 
 const osgi = require('../osgi');
-const utils = require('../utils');
 const log = require('../log')('metadata');
 
 let MetadataRegistry = osgi.getService("org.openhab.core.items.MetadataRegistry");
 let Metadata = Java.type("org.openhab.core.items.Metadata");
 let MetadataKey = Java.type("org.openhab.core.items.MetadataKey");
 
-
 /**
- * This function will return the Metadata object associated with the
- * specified Item.
+ * Class representing an Item's Metadata
  * 
- * @memberof metadata
- * @param {String} name of the Item
- * @param {String} namespace name of the namespace
- * @returns {String|null} the metadata as a string, or null
+ * @memberOf metadata
  */
-let getValue = function(itemName, namespace) {
-    let result = MetadataRegistry.get(new MetadataKey(namespace, itemName));
-    return result ? result.value : null;
-};
+class ItemMetadata {
 
-let addValue = function(itemName, namespace, value) {
-    let key = new MetadataKey(namespace, itemName);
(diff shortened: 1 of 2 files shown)


metadata/metadata.js
```diff
@@ -7,70 +7,113 @@
...
- */
-let upsertValue = function(itemName, namespace, value) {
-    let existing = getValue(itemName, namespace);
+    /**
+    * Adds (inserts) a metadata value.
+    * 
+    * @param {String} namespace the name of the namespace
+    * @param {String} value the value to insert or update
+    */
+    addValue(namespace, value) {
```
marcinczeczko, line 50: I think it's worth to take a full potential of metadata and let the `addValue` method to set also metadata configuration values. currently this method sets it to empty. I was thinking about splitting it into two methods as for the reading configuration values, but apparently the `Metadata` object doesn't have any methods - it's possible only via constructor

metadata/metadata.js
```diff
@@ -7,70 +7,113 @@
...
-    } else {
-        updateValue(itemName, namespace, value);
-        return false;
+    /**
+    * Updates metadata values for this item.
+    * @param {String} namespace The namespace for the metadata to update
+    * @param {String} value the value to update the metadata to
+    * @returns {String} the updated value
+    */
+    updateValue(namespace, value) {
```
marcinczeczko, line 61: If `addValue` would support metadata.configuration, then the update method should do the same

metadata/metadata.js
```diff
@@ -7,70 +7,113 @@
...
+        let metadata = createMetadata(this.itemName, namespace, value);
+        let result = MetadataRegistry.update(metadata);
+        return result ? result.value : null;
+    }
+
+    /**
+     * Updates metadata values for this item.
+     * @param {Map} namespaceToValues A map of namespaces to values to update
+     */
+    updateValues(namespaceToValues) {
```
marcinczeczko, line 71: Here the `namespaceToValues` map should support configuration too. But it's worth to document how to prepare that structure. Wdyt ?

metadata/metadata.js
```diff
@@ -7,70 +7,113 @@
...
+    }
+
+    /**
+     * Adds (inserts) or updates a metadata value.
+     * 
+     * @param {String} namespace the name of the namespace
+     * @param {String} value the value to insert or update
+     * @returns {Boolean} true if the value was added, false if it was updated
+     */
+    upsertValue(namespace, value) {
```
marcinczeczko, line 84: And of course `upsertValue` should handle metadata.configuration too

digitaldan:
Thanks @marcinczeczko for the feedback, I'll take a look at supporting updating the configuration options fully if we are ok with this approach.

PR #44, opened 2021-12-22
JSRule: Overwrite/Update existing rule with same UID or throw error

### LINKED ISSUE #40: Feature request: Updating an existing JSRule()
Is there any way to update an existing JSRule() or to replace it with a new rule?
This would be great to update Triggers, e.g. update a cron trigger.

I would use it in one manager rule to update the triggers of a child rule or I would also be happy with just removing the child rule and then adding a new child rule.

florian-h05:
## Fixes #40 
@jpg0 

On creation of a new rule using `JSRule()`, the `ruleRegistry` is checked for a rule with the same UID. 

When a it already exists, throw error. 
Or:
Use the new `overwrite` parameter to overwrite an existing rule with same UID.
This can be used to dynamically update rules from other rules.

Note: Rule must be created from same file.

### DIFF
--- a/rules/rules.js
+++ b/rules/rules.js
@@ -12,7 +12,7 @@ const utils = require('../utils');
 const log = require('../log')('rules');
 const osgi = require('../osgi');
 const triggers = require('../triggers');
-const { automationManager } = require('@runtime/RuleSupport');
+const { automationManager, ruleRegistry } = require('@runtime/RuleSupport');
 
 let RuleManager = osgi.getService("org.openhab.core.automation.RuleManager");
 let factory = require('@runtime/rules').factory;
@@ -78,6 +78,37 @@ const getGroupsForItem = function (ruleConfig) {
 }
 
 /**
+ * Check whether a rule exists.
+ * Only works for rules created in the same file.
+ *
+ * @memberOf rules
+ * @private
+ * @param {String} uid the UID of the rule
+ * @returns {Boolean} whether the rule exists
+ */
+const ruleExists = function (uid) {
+    return !(ruleRegistry.get(uid) == null);
+};
+
+/**
+ * Remove a rule when it exists. The rule will be immediately removed.
+ * Only works for rules created in the same file.
+ *
+ * @memberOf rules
+ * @param {String} uid the UID of the rule
+ * @returns {Boolean} whether the rule was actually removed
+ */
+const removeRule = function (uid) {
+    if (ruleExists(uid)) {
+        log.info('Removing rule: {}', ruleRegistry.get(uid).name ? ruleRegistry.get(uid).name : uid);
+        ruleRegistry.remove(uid);
+        return !ruleExists(uid);
+    } else {
+        return false;
+    }
+};
+
+/**
  * Creates a rule. The rule will be created and immediately available.
  *
  * @example
@@ -85,7 +116,7 @@ const getGroupsForItem = function (ruleConfig) {
  *
  * rules.JSRule({
  *  name: "my_new_rule",
- *  description": "this rule swizzles the swallows",
+ *  description: "this rule swizzles the swallows",
  *  triggers: triggers.GenericCronTrigger("0 30 16 * * ? *"),
  *  execute: triggerConfig => { //do stuff }
  * });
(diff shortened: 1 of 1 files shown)


rules/rules.js
```diff
@@ -77,8 +77,34 @@ const getGroupsForItem = function (ruleConfig) {
...
+    return !(ruleRegistry.get(uid) == null);
+};
+
+/**
+ * Remove a rule when it exists. The rule will be immediately removed.
+ *
+ * @memberOf rules
+ * @param {String} uid the UID of the rule
+ */
+const removeRule = function (uid) {
```
jpg0, line 98: I would suggest returning a boolean indicating whether a rule was actually removed or not

rules/rules.js
```diff
@@ -96,10 +122,13 @@ const getGroupsForItem = function (ruleConfig) {
  * @param {String} ruleConfig.description a description of the rule
  * @param {*} ruleConfig.execute callback that will be called when the rule fires
  * @param {HostTrigger|HostTrigger[]} ruleConfig.triggers triggers which will define when to fire the rule
+ * @param {String} ruleConfig.id the UID of the rule
+ * @param {String} ruleConfig.ruleGroup the name of the rule group to use.
  * @returns {HostRule} the created rule
  */
 let JSRule = function (ruleConfig) {
     let ruid = ruleConfig.id || ruleConfig.name.replace(/[^\w]/g, "-") + "-" + utils.randomUUID();
+    removeRule(ruid);
```
jpg0, line 131: I'm not sure that it's the best thing to _force_ prior removal. This means that if a script relies on failure if it tries to insert a rule a second time, it will change it's behaviour. More concretely: if two rules with the same uid are inserted, then the existing behaviour is that the second will fail. The new behaviour is that the first will be silently overwritten.

My preference would be to (a) retain the original behaviour by default (as overwriting is not a common use-case, I'd expect that doing it accidentally is more common and failing is good in this case) and (b) allow explicit overwriting by setting a parameter.

jpg0 (changes requested):

rules/rules.js
florian-h05: Two things to mention:

- Removing a rule is only possible when it was created in the same file
- Returning a boolean indicatin whether it was removed or not does not work with:
```javascript
return !(ruleRegistry.remove(uid) == null);
```
It seems that null is also returned when the rule was removed, but accroding to the [Java Doc](https://www.openhab.org/javadoc/latest/org/openhab/core/common/registry/registry#remove(K)) it should return the element that was removed. It really removes the rule, but it always returns `null`.

To return whether the rule was removed, I have to use that work-around:
```javascript
if (ruleExists(uid)) {
   log.info('Removing rule: {}', ruleRegistry.get(uid).name ? ruleRegistry.get(uid).name : uid);
   ruleRegistry.remove(uid);
   return !ruleExists(uid);
} else {
   return false;
}
```

florian-h05:
@jpg0 

I tried to bring in you requested changes. Please have a look at my comment.

rules/rules.js
jpg0: lgtm

jpg0 (approved):

jpg0:
@digitaldan happy to merge this?

rules/rules.js
```diff
@@ -85,21 +116,27 @@ const getGroupsForItem = function (ruleConfig) {
...
  * @param {*} ruleConfig.execute callback that will be called when the rule fires
  * @param {HostTrigger|HostTrigger[]} ruleConfig.triggers triggers which will define when to fire the rule
+ * @param {String} [ruleConfig.id] the UID of the rule
+ * @param {String} [ruleConfig.ruleGroup] the name of the rule group to use
+ * @param {Boolean} [ruleConfig.overwrite=false] overwrite an existing rule with the same UID
  * @returns {HostRule} the created rule
  */
 let JSRule = function (ruleConfig) {
     let ruid = ruleConfig.id || ruleConfig.name.replace(/[^\w]/g, "-") + "-" + utils.randomUUID();
+    if (ruleConfig.overwrite === true) {
```
digitaldan, line 137: So without this option, if a rule exists with the same ID , it would thrown an error when trying to persist it?

florian-h05: No, it would just be quiet without that option set.

digitaldan:
LGTM, i'm assuming I know the answer to the one question i had, so please feel free to merge !

digitaldan (approved):

florian-h05:
> LGTM, i'm assuming I know the answer to the one question i had, so please feel free to merge !

When trying to add a rule with same UID and without the `override` option set, it would be quiet.

I can not merge as I have no write access.

rules/rules.js
digitaldan: huh, thats not actually what i thought would happen :-) is that desired?  i'm thinking we should always then remove the rule if it exists, or thrown an error. So right now, you would have no idea if it save properly or not without this flag right?

florian-h05: I don't think is desired that there is no error, but I am watching this behavious on my system.

I will add something to throw an error as overriding by default is nothing @jpg0 would do.

florian-h05:
@digitaldan 

It now throws an error when a rule with the same UID already exists and the `override` option is not set.

rules/rules.js
```diff
@@ -85,61 +116,71 @@ const getGroupsForItem = function (ruleConfig) {
...
-    if (ruleConfig.name) {
-        rule.setName(ruleConfig.name);
-    }
-
-    //Register rule here
-    if (triggers && triggers.length > 0) {
-        rule.setTriggers(triggers);
-        rule = registerRule(rule);
+    if (ruleExists(ruid)) {
+        log.error('Failed to add rule: {}, a rule with same UID [{}] already exists!', ruleConfig.name ? ruleConfig.name : ruid, ruid);
```
digitaldan, line 141: should we instead throw the error instead of logging it?

florian-h05: Ups, yes, give me a few minutes.

florian-h05:
@digitaldan 

New try :-)

digitaldan:
LGTM, I can merge when I'm back home , unless @jpg0 you want to review and merge ?

jpg0:
Ideally `ScriptedRuleProvider` in core should return whether a rule deletion actually occurs (instead of @florian-h05 needed to perform another check after attempting the removal), but that's too much to fix right now, so let's merge this.

PR #46, opened 2021-12-23
Optimized version to also build TypeDefinition files (Visual Studio Code and TypeScript support)

csowada:
This is my fork to create d.ts files for the project. I've also changed some imports. Please let us discuss if it goes into the right direction or not.


digitaldan:
Thanks @csowada !  Yes i think this is the right direction. I also played around with the same strategy using  tsc to create the types and agree the code needed some moving around.  One thing to keep in mind is our API documentation, JSDoc does not like the Typescript specific `typedef` syntax and blows up trying to parse it.  Try running `npm run docs` to see what i'm talking about.  Not sure what to do about that.

csowada:
Uff, I was not aware of the jsdoc process. Maybe it is faster to create a TS port and build a native JS version. It also helps for better code, more strict etc. Afterward we could apply jsdoc on the JS output

PR #51, opened 2021-12-31
Monkey Patches ZonedDateTime.parse to default to OH time format

### LINKED ISSUE #45: Issue with joda-js ZonedDateTime parser - default format not compatible with DateTimeItem state value
I was working on the rule that uses openhab `DateTimeItem` state to get the time to schedule some timers. 
However, I found strange issue that parsing the state string as it is throws joda parsing exception. I had to instantiate my own formatter to get it working. Please take a look below, but in general now idea if the problem lies in the openhab datetime format, joda-js default format or maybe oh-js itself.

### how to reproduce
- In the rule, or script just do this
```js
let tm = "2021-12-22T19:08:05.123+0100";
console.log(time.ZonedDateTime.parse(tm));
```
An exception is being throws as follows
```
19:14:29.215 [ERROR] [.internal.handler.ScriptActionHandler] - Script execution of rule with UID 'sun_state_rule' failed: org.graalvm.polyglot.PolyglotException: JsJodaException: Text '2021-12-22T19:08:05+0100' could not be parsed at index 19: 2021-12-22T19:08:05+0100, at index: 19
```
### workaround
- As a workaround I simply implemented the parsing method to have it parsed.
```
function toJoda(itemDateTime) {
      const df = time.DateTimeFormatter.ofPattern("yyyy-MM-dd'T'HH:mm:ss.SSSZ");
      return time.ZonedDateTime.parse(itemDateTime, df);
}
```
### findings
Didn't get much time looking at the defaults in joda-js, but looks like it expects timezone in the form of `+01:00` with colon symbol, while openhab reports with it, e.g. `+0100`. So, these is parsed correctly
```js
let tm = "2021-12-22T19:08:05.123+01:00";
console.log(time.ZonedDateTime.parse(tm));
```

digitaldan:
Fixes #45

### DIFF
--- a/index.js
+++ b/index.js
@@ -18,5 +18,5 @@ module.exports = {
     get utils() { return require('./utils') },
     get osgi() { return require('./osgi') },
     get cache() { return require('./cache') },
-    get time() { return require('@js-joda/core') }
+    get time() { return require('./time') }
 }
--- a/time.js
+++ b/time.js
@@ -1,0 +1,12 @@
+
+const time = require('@js-joda/core');
+
+//openHAB uses a RFC DateTime string, js-joda defaults to the ISO version, this defaults RFC instead
+const rfcFormatter = time.DateTimeFormatter.ofPattern("uuuu-MM-dd'T'HH:mm:ss.SSS[xxxx][xxxxx]");
+const targetParse = time.ZonedDateTime.prototype.parse;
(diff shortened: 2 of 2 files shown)


digitaldan:
If there are no objections (or alternative ideas to plug in a default parser), then i will merge.

PR #60, opened 2022-01-07
Fix create of Trigger with old state (from)

Hilbrand:
The configuration parameter in the StateChangeTrigger is named 'previousState' and not 'oldState'.
'oldState' is the value in the event, but should not be used to construct the trigger.
See https://github.com/openhab/openhab-core/blob/cabb3f73150d13f44d647f46af8362e396684b92/bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/internal/module/handler/ItemStateTriggerHandler.java#L67

### DIFF
--- a/triggers.js
+++ b/triggers.js
@@ -68,7 +68,7 @@ module.exports = {
     ItemStateChangeTrigger: (itemName, oldState, newState, triggerName) => createTrigger("core.ItemStateChangeTrigger", triggerName, {
         "itemName": itemName,
         "state": newState,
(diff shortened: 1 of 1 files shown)


digitaldan:
Thanks @Hilbrand !  Question, should we also rename the `oldstate` variable name in the method signature just for clarity?

Hilbrand:
I thought about it. But given the other is called `newState` it seems more logic to keep it `oldState`.

digitaldan:
Thanks!

PR #61, opened 2022-01-07
Update rules.js

lukicsl:
Return also full content of input map data. This is useful to get the triggerName when it is specified during trigger definition (module tag)

### DIFF
--- a/rules/rules.js
+++ b/rules/rules.js
@@ -275,7 +275,8 @@ const getTriggeredData = function (input) {
             receivedCommand: event.getItemCommand(),
             oldState: input.get("oldState") + "",
             newState: input.get("newState") + "",
-            itemName: event.getItemName()
+            itemName: event.getItemName(),
+            input: input
         }
     }
 
@@ -298,6 +299,7 @@ const getTriggeredData = function (input) {
         receivedState: null,
         receivedTrigger: null,
         itemName: evArr[0]
+        input: input
     };
 
     try {


digitaldan:
Thanks for the PR!  @jpg0 this brings up a question i was thinking about.  I think GraalVM by default does not try and convert Java Maps to native JS objects by default?  So `input` here will be a Java Map, correct?  Is it worth while to convert Maps -> JS Objects ?  @lukicsl what are the input values in this map?  I'm a little embarrassed that i'm not confident of their types.

lukicsl:
> what are the input values in this map? I'm a little embarrassed that I'm not confident of their types.

What impact does it have not being confident about the types?
for a channel event trigger it looked like this (log output), implying these are strings:

`{OPT1.event=homematic:HmIP-SMI55:3014F711A0001F5A4993FA84:0014D709AEF7C0:1#BUTTON triggered SHORT_PRESSED, event=homematic:HmIP-SMI55:3014F711A0001F5A4993FA84:0014D709AEF7C0:1#BUTTON triggered SHORT_PRESSED, module=OPT1}
`
The trigger definition itself is:

`triggers.ChannelEventTrigger('homematic:HmIP-SMI55:3014F711A0001F5A4993FA84:0014D709AEF7C0:1#BUTTON', 'SHORT_PRESSED', "OPT1")`

jpg0:
>  I think GraalVM by default does not try and convert Java Maps to native JS objects by default? So input here will be a Java Map, correct? Is it worth while to convert Maps -> JS Objects ?

Correct. If they are Java objects you lose all the JS syntax like for loops etc, and the methods are annoying similar but different. Same for Java Sets.

lukicsl:
@digitaldan  I am new to PR stuff. Is any action needed from me to bring this PR workflow to the next step?

lukicsl:
Will reissue more concrete PR

digitaldan:
Hi @lukicsl ,  for future reference you could always keep this open and just push changes to it.  I am just finding time today to review openHAB related work, hence the delay.  What i was contemplating was creating a helper function in the utils class to convert Java Maps to JS Objects.  We want to avoid exposing Java native stuff as much as we can, and even though the `input` Map in Java can have Java native values, i think they should always be Strings, so converting this to a JS objects seems like a good approach.

lukicsl:
> for future reference you could always keep this open and just push changes to it
I am new to the PR workflow, so I created the new PR#66. It is more concrete in the intention.

PR #65, opened 2022-01-08
Rules: Add ability to define tags

### LINKED ISSUE #57: Add ability to define Rule tags
Currently neither JSRule nor RuleBuilder doesn't support adding Tags to the create rule. Maybe there is a reason why it is not available that I'm not aware. Anyway, I can contribute with this addition if necessary.

florian-h05:
@marcinczeczko
Adds the ability to define rule tags in both `JSRule` and `RuleBuilder` syntax.
Tags must be an Array of Strings, in the `RuleBuilder` syntax the tag `js-rule-builder` is always added.

I also updated the README for the tags, completed the Table of Contents and fixed the link to the JS-Joda docs.

Fixes #57.

### DIFF
--- a/rules/operation-builder.js
+++ b/rules/operation-builder.js
@@ -17,23 +17,25 @@ class OperationBuilder {
         }
     }
 
-    _then(operation, group, name, description) {
+    _then(operation, group, name, description, tags) {
         this.builder.name = name;
         this.builder.description = description;
+        this.builder.tags = tags;
         this.builder.setOperation(operation, group);
     }
 
     /**
      * Build this rule
      *
-     * @param {string} name of the rules
-     * @param {string} description of the rule
+     * @param {string} [name] of the rule
+     * @param {string} [description] of the rule
+     * @param {Array<String>} [tags] of the rule
      */
-    build(name, description) {
+    build(name, description, tags) {
         if (!this.fn) {
             throw new Error("Cannot call build without function")
         }
-        this._then(this.fn, this.group, name, description);
+        this._then(this.fn, this.group, name, description, tags);
     }
 
     /**
@@ -171,11 +173,12 @@ class OperationConfig {
     /**
      * Build this rule
      *
-     * @param {string} name of the rules
-     * @param {string} description of the rule
+     * @param {string} [name] of the rule
+     * @param {string} [description] of the rule
+     * @param {Array<String>} [tags] of the rule
      */
-    build(name, description) {
-        this.operationBuilder._then(this, this.group, name, description);
+    build(name, description, tags) {
+        this.operationBuilder._then(this, this.group, name, description, tags);
     }
 }
 /**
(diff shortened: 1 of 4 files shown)


Confectrician:
Why did you remove the capitalization in so much (sub)headlines?
Capitalization is a common style practice for docs healdines, even if the terms should/must be used in the code without it.

florian-h05:
@Confectrician 
I thought that the headlines should be the same way in the headline and the table of contents, therefore I removed the capitalization, the namespaces are also written not capitalized.

I can revert back to capitalized headlines, but then I would like to capitalize the entries in the table of contents so that they are the same way.

Confectrician:
From a reading persepective (and that's what the readme aims for) capitalization is the common way for handling the headings, so i would vote for capitalization.

As long as the code examples are not capitalized (only TOC and Headings) this should be fine.
@digitaldan do you have any problems with this?

florian-h05:
@Confectrician 

I have capitalized all headings and I have changed the level of heading for the subheadings of `Actions`, therefore subheadings like `Exec` are not displayed in the ToC anymore.

rules/rule-builder.js
```diff
@@ -89,7 +89,8 @@ class RuleBuilder {
             ruleGroup: optionalRuleGroup,
             execute: function (data) {
                 fnToExecute(data);
-            }
+            },
+            tags: this.tags || ['js-rule-builder']
```
digitaldan, line 93: Why set a default tag here?  I would assume if no tags are supplied, we should not add one ?

rules/operation-builder.js
```diff
@@ -17,23 +17,26 @@ class OperationBuilder {
         }
     }
 
-    _then(operation, group, name, description) {
+    _then(operation, group, name, description, tags) {
         this.builder.name = name;
         this.builder.description = description;
+        if (tags) tags.push('js-rule-builder');
```
digitaldan, line 23: Why add a default tag here as well?

digitaldan:
Thanks @florian-h05 for the PR!  I have 2 small comments, but otherwise this looks good.

digitaldan (changes requested):
Please see my comments about default tags being added.

rules/rule-builder.js
florian-h05: I thought having the `js-rule-builder` tag always added would be helpful, as you would know where a rule comes from in the UI.

florian-h05:
Should I remove the default tags?

florian-h05:
@digitaldan 

I have removed the default tag. Ready for your review.

florian-h05:
@digitaldan ?

digitaldan:
Hi , was not able to work on openHAB at all this week, i try and pull down all PR's and test them before merging.  I'm going to do that later today, thanks.

digitaldan (approved):
LGTM

PR #66, opened 2022-01-09
getTriggeredData shall deliver "module" (triggerName) 

lukicsl:
getTriggeredData shall return also the module name, (the trigger name supplied at creation time)

This would make it possible to have supply names to the triggers, e.g.:

```
trig.push(triggers.GenericCronTrigger("0 0/1 * 1/1 * ? *", "CRON"))
trig.push(triggers.ChannelEventTrigger('homematic:HmIP-SMI55:3014F711A0001F5A4993FA84:0014D709AEF7C0:1#BUTTON', 'SHORT_PRESSED', "OPT1"))
trig.push(triggers.ChannelEventTrigger('homematic:HmIP-SMI55:3014F711A0001F5A4993FA84:0014D709AEF7C0:2#BUTTON', 'SHORT_PRESSED', "OPT2"))

```
and the easy do evaluate at the execute which triggers is executed:

```
rules.JSRule({
  name: "Test Rule",
  description: "Firste test for migration from Jython to JS",
  triggers: trig,
  execute: module => {
    switch (module["module"]) {
        case "CRON":
             ...
            break;
        case "OPT1":
             ...
            break;
        case "OPT2":
             ...
            break;
    }

  }
});

```

### DIFF
--- a/rules/rules.js
+++ b/rules/rules.js
@@ -275,7 +275,8 @@ const getTriggeredData = function (input) {
             receivedCommand: event.getItemCommand(),
             oldState: input.get("oldState") + "",
             newState: input.get("newState") + "",
-            itemName: event.getItemName()
+            itemName: event.getItemName(),
+            module: input.get("module")
         }
     }
 
@@ -297,7 +298,8 @@ const getTriggeredData = function (input) {
         receivedCommand: null,
         receivedState: null,
         receivedTrigger: null,
-        itemName: evArr[0]
+        itemName: evArr[0],
+        module: input.get("module")
     };
 
     try {
--- a/rules/rules.js
+++ b/rules/rules.js
@@ -336,6 +336,7 @@ const getTriggeredData = function (input) {
                 d.eventType = "time";
                 d.triggerType = "GenericCronTrigger";
                 d.triggerTypeOld = "TimerTrigger";
+                d.module = "CRON";
             } else {
                 d.eventType = "";
                 d.triggerType = "";
--- a/rules/rules.js
+++ b/rules/rules.js
@@ -332,11 +332,11 @@ const getTriggeredData = function (input) {
             d.receivedTrigger = evArr[2];
             break;
         default:
+            d.module = "";
             if (input.size() == 0) {
                 d.eventType = "time";
                 d.triggerType = "GenericCronTrigger";
                 d.triggerTypeOld = "TimerTrigger";
-                d.module = "CRON";
             } else {
                 d.eventType = "";
                 d.triggerType = "";
--- a/rules/rules.js
+++ b/rules/rules.js
@@ -332,7 +332,6 @@ const getTriggeredData = function (input) {
             d.receivedTrigger = evArr[2];
             break;
         default:
-            d.module = "";
             if (input.size() == 0) {
                 d.eventType = "time";
                 d.triggerType = "GenericCronTrigger";


rules/rules.js
```diff
@@ -334,6 +336,7 @@ const getTriggeredData = function (input) {
                 d.eventType = "time";
                 d.triggerType = "GenericCronTrigger";
                 d.triggerTypeOld = "TimerTrigger";
+                d.module = "CRON";
```
digitaldan, line 339: Whats the purpose of injecting this here?

rules/rules.js
```diff
@@ -297,7 +298,8 @@ const getTriggeredData = function (input) {
         receivedCommand: null,
         receivedState: null,
         receivedTrigger: null,
-        itemName: evArr[0]
+        itemName: evArr[0],
+        module: input.get("module")
```
digitaldan, line 302: This looks ok to me , although this will return a null value if not found, i think thats ok as long as we don't have any code that expects otherwise (like i'm assuming `oldState` and `newState` where we set to a empty string)   i'm not familiar with the module in the `input` , looks like its an ID/UUID to something , out of curiosity what are you using it for in rules ?

digitaldan (changes requested):

rules/rules.js
lukicsl: > out of curiosity what are you using it for in rules

see above in the initial description.

for simple rules, I  implemented my own design pattern as described above, where I supply names for triggers, e.g. "BT1", "BT2", these labels I then use in a switch statement to execute code based on which trigger happened.

I also have some other rules (lights, shades, ..) where a single rule might have 100+ triggers built out of a JSON definition string. see: [https://community.openhab.org/t/tutorial-example-of-lights-control-using-lambdas-ported-to-ngre-python/68601?u=lukics](https://community.openhab.org/t/tutorial-example-of-lights-control-using-lambdas-ported-to-ngre-python/68601?u=lukics).

My rules are right now all written in Jython. I am preparing the ground to port these to JS.

rules/rules.js
lukicsl: for CRON type of triggers, the input is empty.  It would be also totally acceptable to have d.module an empty string.

rules/rules.js
```diff
@@ -330,6 +332,7 @@ const getTriggeredData = function (input) {
             d.receivedTrigger = evArr[2];
             break;
         default:
+            d.module = "";
```
digitaldan, line 335: why are you setting this here? Isn't this already defined above ?

lukicsl: for cron triggers input.get("module") is undefined, so setting it here too

digitaldan: i'm guessing you mean it will be null, not undefined?  Is that a problem?  I think null would be better then an empty string ?

lukicsl: True, let's get rid of this change

digitaldan (approved):
LGTM

lukicsl:
@digitaldan I have one of the checks not passed (DCO). I have meanwhile signed off the commits, as suggested in the help text. No clue what to do else.

digitaldan:
The PR was merged and is part of the main branch, so the next release we tag and push to npm will include your change.  For future reference, you should sign your commits by adding the `-s` flag, for example `git commit -as` .  As this change was small, I bypassed the check.

PR #70, opened 2022-01-15
Add item.getItems()

### LINKED ISSUE #56: Add ability to retrieve all items
Dear community,

it would be great if its possible to retrieve all items using jsscripting. See https://www.openhab.org/javadoc/latest/org/openhab/core/items/itemregistry#getItems()

For no I am using a workaround.

digitaldan:
Fixes #56

### DIFF
--- a/README.md
+++ b/README.md
@@ -200,6 +200,7 @@ See [openhab-js : items ](https://openhab.github.io/openhab-js/items.html) for f
 
 * items : <code>object</code>
     * .getItem(name, nullIfMissing) ⇒ <code>Item</code>
+    * .getItems() ⇒ <code>Array.&lt;Item&gt;</code>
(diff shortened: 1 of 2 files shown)


jpg0:
Maybe `getItemsByTag(..)` should be replaced by this method?

E.g. `getItems(tags = [], type = null)` etc, so that we can use a single method with filters rather than multiple specific methods?

digitaldan:
I agree that seems nicer, although since the item registry  has a `getItemsByTag` method, i think its ok to mirror this here ?  There is also the issue of breaking changes (which i'm adding to the README) , although we could mark this as depreciated and remove in a future release?  I'm ok either way if you have a strong preference?

jpg0:
Let's ignore it for the moment, and revisit if we get a request for more `getItemByXXX` methods.

PR #71, opened 2022-01-15
Adds rule ID to rulebuilder

### LINKED ISSUE #68: How to run/execute js rules from ui over long term with restarts and changes
Currently the js rules produce an identifier based on the rule name and a generated uid.
The uid changes on every rule file save and when openHAB is restarted. (Not sure about the last one tough.)

image

Configuration yaml code for the actions:

```yaml
actions:
  - inputs: {}
    id: "4"
    configuration:
      considerConditions: true
      ruleUIDs:
        - Wake-Up-Radio-Automatic-c5b258d4-b32c-43d0-a38d-1f34ab8fa7bd
    type: core.RunRuleAction
    label: Start Wake Up Radio
  - id: "5"
    configuration:
      enable: "false"
      ruleUIDs:
        - arriveAtHome
    type: core.RuleEnablementAction
    label: Disable after first run
```

When i want to run/execute a JS rule from another ui rule, this will only work until the system got restarted or i have changed something.
To get it working again, i have to adapt all rules that are used for running/executing js rules to get the current name with the newly generated uid.

My current workaround would be using virtual items as triggers in the js rules, but this is not very intuitive.
Does someone know a better solution to this?

Probably this is more relevant for openhab-core because it may affect other scripting languages too.

digitaldan:
Fixes #68

### DIFF
--- a/README.md
+++ b/README.md
@@ -532,8 +532,10 @@ Rules are started by calling `rules.when()` and can chain together [triggers](#r
 [conditions](#rule-builder-conditions) and [operations](#rule-builder-operations) in the following pattern:
 
 ```javascript
-rules.when().triggerType()...if().conditionType().then().operationType()...build(name, description, tags);
+rules.when().triggerType()...if().conditionType().then().operationType()...build(name, description, tags, id);
 ```
+
+Note: `name`, `description`, `tags` and `id` are optional and reasonable defaults will be used if omitted.
 
 Rule are completed by calling `.build(name, description, tags)` , if name or description are omitted, a generated value will be used.
 If tags is used, it must be an Array of strings.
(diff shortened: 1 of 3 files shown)


jpg0:
Does this _require_ an ID, or is it optional?

florian-h05:
I think it is optional, as `RuleBuilder` is utilizing `JSRule` under the hood and for `JSRule` the id is optional.
But maybe somebody could verify this?

digitaldan:
It optional,  which is stated in the JSDocs, I also added a line to the readme stating that and added some examples that show both with and without supplying an ID.  I think this is clear enough ?

digitaldan:
I did realize i missed one more section in the README in the rule builder section, i'll update that as well.

jpg0:
Yes, I think that is fine. I was just asked as I was a little worried about it being mandatory, and also the proliferation of optional parameters on the `build` method (rather than being methods themselves, like `id(..)`). Probably fine with id though as it's sufficiently core to the object.

PR #78, opened 2022-01-19
utils.js: Add documentation

### LINKED ISSUE #75: utils.js missing from generated docs
The methods in utils are quite handy. We should make sure they get into the docs too. Even if just in https://openhab.github.io/openhab-js/index.html. 

I've no idea how the docs are generated (it's on my list of things to learn) or I'd just submit a PR right now. I'll do it when I can figure it out but if it's just a quick change don't wait for me.

florian-h05:
Fixes #75.

@rkoshak 
I mentioned the utils in the README.
I have added most of the functions to the JSDoc.

Currently, the JSDoc for utils looks like (the screenshot only shows the beginning):
utils JSDoc Beginning
I have included links to the docs for the Java types and the JavaScript Set Object.

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,6 +38,7 @@ binding](https://www.openhab.org/addons/automation/jsscripting/)
   - [Cache](#cache)
   - [Log](#log)
   - [Time](#time)
+  - [Utils](#utils)
 - [File Based Rules](#file-based-rules)
   - [JSRule](#jsrule)
   - [Rule Builder](#rule-builder)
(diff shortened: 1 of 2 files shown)


rkoshak:
`dumpObject` looks like it operates on anything so I guess it would by `*`. The purpose appears to be to log out all the members and properties and whether it's a JS Object or a Java Object. It doesn't return anything but it does throw an error if a problem is encountered.

The same for `instance` on `isJsInstanceOfJava`. `type` on the function I think would be a Java `java.lang.Class`. It returns a boolean if the Object is a Java Object as opposed to a pure JavaScript Object.

PR #80, opened 2022-01-20
Fix throw Error

### LINKED ISSUE #79: 'throw error' causes: ReferenceError: "error" is not defined
When playing around with openhab-js, I experienced an issue, where the JS code should throw an error, but instead GraalJS throws:
```
2022-01-20 18:25:20.009 [ERROR] [internal.handler.ScriptActionHandler] - Script execution of rule with UID 'test' failed: org.graalvm.polyglot.PolyglotException: ReferenceError: "error" is not defined
```

Please note: While opening this issue, I already have a fix. I only opened this issue for better bug tracking.

florian-h05:
Fixes #79.

Instead of `throw Error` sometimes `throw error` was used,
which led to reference errors.

### DIFF
--- a/rules/operation-builder.js
+++ b/rules/operation-builder.js
@@ -417,7 +417,7 @@ class ToggleOperation extends OperationConfig {
                 break;
             }
             default:
-                throw error(`Toggle not supported for items of type ${item.type}`);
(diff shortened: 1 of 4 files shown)


digitaldan:
good catch !  (pun intended)

PR #81, opened 2022-01-20
Added runRule, isEnabled and setEnabled

rkoshak:
I also changed 'ruleExists` as the RuleManager seems to be more reliable in UI rules than `ruleRegistry`. Sometimes `ruleRegistry` would say a rule doesn't exist when it does buy `RuleManager` always returns a rule's status.

Errors are thrown if the rule doesn't exist in all of these. An error is also thrown if `runRule` is called and the rule is UNINITIALIZED which can be because it's been disabled or is broken somehow).

Signed-off-by: Richard Koshak <rlkoshak@gmail.com>


rules/rules.js
```diff
@@ -108,6 +108,55 @@ const removeRule = function (uid) {
...
+ * 
+ * @memberOf rules
+ * @param {String} uid the UID of the rule to run
+ * @param {Map<Object>} args optional dict of data to pass to the called rule
+ * @param {Boolean} cond when true, the called rull will only run if it's conditions are met
+ */
+const runRule = function (uid, args = {}, cond = true) {
+    const status = RuleManager.getStatus(uid);
+    if(!status) {
+        throw 'There is no rule with UID ' + uid;
```
florian-h05, line 123: May we use `throw Error()` instead of `throw ()` for consistency across openhab-js?

rules/rules.js
```diff
@@ -108,6 +108,55 @@ const removeRule = function (uid) {
     }
 };
 
+/**
+ * Runs the rule with the given UID. Throws errors when the rule doesn't exist
+ * or is unable to run (e.g. it's disabled);
```
florian-h05, line 113: I think the the semicolon is a typo and should be a point.

rules/rules.js
```diff
@@ -108,6 +108,55 @@ const removeRule = function (uid) {
...
+ * 
+ * @memberOf rules
+ * @param {String} uid the UID of the rule to run
+ * @param {Map<Object>} args optional dict of data to pass to the called rule
+ * @param {Boolean} cond when true, the called rull will only run if it's conditions are met
```
florian-h05, line 118: Typo: should be rule and not rull

florian-h05: `args` and `cond` are optional and have defaults.
You can adjust the JSDoc to show that:
```js
/** 
 * …
 * @param {Map<Object>} [args={}] optional dict …
 * @param {Boolean} [cond=true] when true, …
 */
```

florian-h05 (changes requested):
@rkoshak 
Overall looks good, just a few points.
Please have a look at my comments.

rules/rules.js
rkoshak: Note that 109 is not among my changes. That was there to begin with.

I do notice it's not consistent in how it's used in this file so I can change them all to be consistent.

I don't think `.` is right though. It really is the end of a statement, the assignment of the function to `removeRule`.

rules/rules.js
rkoshak: I'll fix it.

rules/rules.js
florian-h05: I think we are not talking about the same ;-)

> I don't think . is right though. It really is the end of a statement, the assignment of the function to removeRule.

That’s correct.

I mean the semicolon at the end of the comment in line 113.

rules/rules.js
rkoshak: Done

rules/rules.js
rkoshak: Done

rkoshak:
The three changes requested have been made. Pending any other comments this PR is good to go.

rules/rules.js
florian-h05: Did you fix that? I can‘t see changes.

florian-h05:
Sign-Offs are missing, I don‘t know how important they are.

Maybe you can add the missing signoffs to the commits or revert the commits till the first one, commit again in one commit with all new changes and then force push (that‘s my way of going when I want to fix something in a previous commit).

rkoshak:
I added one to the most recent commit but researching how to add it to the one before that.

Confectrician:
There is an explanation on the checks page for what is wrong and how to solve it:
https://github.com/openhab/openhab-js/pull/81/checks

rkoshak:
I tried the `git rebase --signoff HEAD~4` followed by the `git push --force-with-lease origin runRule` but it doesn't appear to have kicked off the DCO checks again. Or it didn't work, though I did get "Successfully rebased and updated" on my end.

rkoshak:
Well none of that worked. Guess I'll try reverting and starting over.

rkoshak:
This is all so now hopelessly broken I'll just start over with a new PR. It let me revert a couple of commits but it won't let me do it for others. Now I'm getting errors on pushes and pulls. New PR is #82. Closing this one. All the above comments are addressed in #82.

Sorry for the confusion. Don't commit before your first cup of coffee is the lesson I've learned.

Issue #77, opened 2022-01-19
add runRule function to rules namespace


rkoshak:
Recent questions on the forum indicate there is a strong demand to create rules that call other rules. This is not directly supported by the library. Before I submit a PR I want to make sure this seems like something appropriate for openhab-js.

To me it seems like an obvious addition but will keep it in my own library if that isn't a direction we want to go. The code it pretty simple and it supports running a rule by UID or it will go and search for the rule by name. But I don't know if the search by name part will work in files for rules outside the one file, similar to the limitation with removeRule and findRule. I need to test that.

For reference:

```javascript
const runRule = (nameOrUid, argsDict, cond = false) => {
  // If it's not a UID, try to find it by name
  if(!RuleManager.getStatusInfo(nameOrUid)) {
    const rule = utils.javaSetToJsArray(ruleRegistry.getAll())
                      .find(rule => rule.getName(nameOrUid));
    nameOrUid = (rule) ? rule.getUID() : null;
  }

  // run the rule
  if(nameOrUid) {
    RuleManager.runNow(nameOrUid, cond, argsDict);
    return true;
  }
  else {
    return false;
  }
}
```

digitaldan:
Sounds reasonable, question, can the rule manager run a disabled rule, so should that be checked first?  Also would it be better to throw an error if the rule can not be found or is disabled ?

Confectrician:
Since we have the ability to define fixed and stable rule uids in ui and files, wouldn't it be good to run Rules only from those known id's instead of trying to guess something?

From my view the needed effort for finding the correct uid could be tolerated for a stable function.

> Also would it be better to throw an error if the rule can not be found or is disabled ?

To make it a bit more complicated (😜) we could also think about alowwing to run disabled rules via an additional parameter, set to `false` in default.

rkoshak:
If the rule is disabled the RuleEngine will log out an error.

    2022-01-19 13:52:53.142 [ERROR] [e.automation.internal.RuleEngineImpl] - Failed to execute rule ‘tocall' with status 'UNINITIALIZED'

That's why I opted for a true/false return value. It's a little more informative than what OH itself does without being as disruptive as throwing an error. But I can absolutely check that the rule exists first and I'm fine with throwing an error in cases where it's disabled or can not be found. 

It's not clear to me what RuleManager does if you try to ask it to run a rule that doesn't exist. I'll have to mock up a quick test...

It also just logs a warning, this time

[...]

PR #82, opened 2022-01-21
Run rule2

rkoshak:
Address #77 . Second attempt at this PR to fix DCO errors. Replaces PR #81.


florian-h05 (changes requested):
Two suggestions for the JSDoc comment of runRule.
Please have a look at my comments.

Apart from that it looks good to me :-).

florian-h05:
@rkoshak: May it helps you

When you use VS Code and you have installed the git extension, you can enable `Always Sign Off` under `Settings` -> `Extensions` -> `Git`.

And with the sidebar‘s Git tab, you can revert the last commit by pressing the three points (in the upper right corner of that tab) -> Commit.
In the filed for the commit message, you can use arrow up/down to reuse a commit message. After committing a reverted commit, you MUST NOT press the sychronize button, instead open the Terminal inside VS Code and force-push (e.g. `git push origin --force`).

rules/rules.js
```diff
@@ -108,6 +108,56 @@ const removeRule = function (uid) {
...
 };
 
+
+/**
+ * Runs the rule with the given UID. Throws errors when the rule doesn't exist
+ * or is unable to run (e.g. it's disabled).
+ * 
+ * @memberOf rules
+ * @param {String} uid the UID of the rule to run
+ * @param {Map<Object>} args optional dict of data to pass to the called rule
```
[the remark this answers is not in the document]
rkoshak: Done

rules/rules.js
```diff
@@ -108,6 +108,56 @@ const removeRule = function (uid) {
...
+ * @param {Boolean} cond when true, the called rule will only run if it's conditions are met
```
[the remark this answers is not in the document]
rkoshak: Done

rkoshak:
Thanks for the tips! I've changed the setting so hopefully won't run into this problem again, as long as I'm using VSCode at least.

florian-h05 (approved):
LGTM!

For anyone watching: I have tested the changes.

Confectrician:
Just as a sidenote @rkoshak

If you are using vscode you can enable signing off comits as default, evenb if you are working with the git sidebar in vscode.

THe setting is
```
"git.alwaysSignOff": true
```

You just have to configure the proper sign off credentials in git.
(Which you probably have done already for a manual sign off via console commit already anyway.)

rules/rules.js
```diff
@@ -87,7 +87,7 @@ const getGroupsForItem = function (ruleConfig) {
  * @returns {Boolean} whether the rule exists
  */
 const ruleExists = function (uid) {
-    return !(ruleRegistry.get(uid) == null);
+    return !(RuleManager.getStatusInfo(uid) == null);
```
digitaldan, line 90: Out of curiosity, whats the benefit of changing this line ?

digitaldan:
Thanks @rkoshak this looks good, i only have one comment.  And thanks @florian-h05 for testing.

rules/rules.js
```diff
@@ -108,6 +108,56 @@ const removeRule = function (uid) {
...
 
+
+/**
+ * Runs the rule with the given UID. Throws errors when the rule doesn't exist
+ * or is unable to run (e.g. it's disabled).
+ * 
+ * @memberOf rules
+ * @param {String} uid the UID of the rule to run
+ * @param {Map<Object>} [args={}] args optional dict of data to pass to the called rule
+ * @param {Boolean} [cond=true] when true, the called rule will only run if it's conditions are met
```
digitaldan, line 119: should we add a throws line as well?  for example
```javascript
@throws Will throw an error if the rule is not found or is disabled
```

and also in the two functions below ? 

https://jsdoc.app/tags-throws.html

rules/rules.js
rkoshak: I think I lost that comment. In UI rules, using the ruleRegistry doesn't work for some reason after the rule is run the second time around. The RuleRegistry gets stale somehow. But the RuleManager works every time.

rules/rules.js
rkoshak: I started to but it's not on any of the other methods that throw Errors in this file so I stuck with that as the convention. Should I add @throws to them all or just mine?

digitaldan: Thanks!

digitaldan (approved):
LGTM

PR #90, opened 2022-02-06
Cache: Check whether key exists

florian-h05:
Added a method to cache to check whether a key-value map exists.

Also changed from `let` to `const` as I don't see a reason it has to be `let`.

### DIFF
--- a/cache.js
+++ b/cache.js
@@ -5,7 +5,7 @@
  * @namespace cache
  */
 
-const cache = require("@runtime").sharedcache
+const cache = require("@runtime").sharedcache;
 
 /**
  * Returns the value to which the specified key is mapped
@@ -27,13 +27,13 @@ const cache = require("@runtime").sharedcache
  * @param {function} [defaultSupplier] if the specified key is not already associated with a value, this function will return a default value
  * @returns {(*|null)} the current object for the supplied key, a default value if defaultSupplier is provided, or null
  */
-let get = function (key, defaultSupplier) {
+const get = function (key, defaultSupplier) {
     if (typeof defaultSupplier === 'function') {
         return cache.get(key, defaultSupplier);
     } else {
         return cache.get(key);
     }
-}
+};
 
 /**
  * Associates the specified value with the specified key
@@ -43,9 +43,9 @@ let get = function (key, defaultSupplier) {
  * @param {*} value value to be associated with the specified key
  * @returns {(*|null)} the previous value associated with null, or null if there was no mapping for key
  */
-let put = function (key, value) {
+const put = function (key, value) {
     return cache.put(key, value);
-}
+};
 
 /**
  * Removes the mapping for a key from this map if it is present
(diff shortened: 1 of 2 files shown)


cache.js
```diff
@@ -54,12 +54,25 @@ let put = function (key, value) {
...
 
+/**
+ * Checks whether the mapping for a key from this map exists.
+ * 
+ * @memberof cache
+ * @param {string} key key whose mapping is to be checked in the map
+ * @returns {boolean} whether the key has a mapping therefore whether it exists
+ */
+const exists = function (key) {
+    return (!get(key, false) === false) ? true : false;
```
digitaldan, line 69: The corner case here would be if the cache value is a boolean, and is set to false, then calling `exist` would always return false ? Maybe this should use `null` instead of a boolean in the default supplier ?

digitaldan: err scratch my last comment, i'm not sure i'm  following the `!get(key, false) === false` logic.  `get` would return null here if the key does not exist?  Whats the benefit of calling `get(key, false)` over just `get(key)`?

florian-h05: There is no benefit, I just did not think about the null option.
I have updated my PR.

florian-h05:
@digitaldan 
I have updated the PR.

Out of interest:

- Could we use some linter for consistent code-style? I am thinking of something like [standard/semistandard](https://github.com/standard/semistandard).
- Should we use a `.npmignore` file do exclude e.g. images from the npm package?

cache.js
```diff
@@ -54,12 +54,25 @@ let put = function (key, value) {
...
+
+/**
+ * Checks whether the mapping for a key from this map exists.
+ * 
+ * @memberof cache
+ * @param {string} key key whose mapping is to be checked in the map
+ * @returns {boolean} whether the key has a mapping therefore whether it exists
+ */
+const exists = function (key) {
+    return (get(key) === null) ? false : true;
```
digitaldan, line 69: wouldn't `return get(key) !== null` suffice to return ?

florian-h05: Yes, it is the same (personally, I prefer the ternary if as it is more clear for me).

digitaldan (approved):
LGTM, thanks.

jpg0:
I'm late to this, and it may not matter, but technically you can store nulls in the cache as markers - at least that is supported in the underlying Java map. To mirror the Java functionality you would call through to the `contains` method on the underlying Java map.

florian-h05:
Okay, currently I have no idea why someone should store null in the map, but I will have a look at using `contains`. When I get it working, I will open a new PR to fix that.

jpg0:
Yes, it's a bit weird. I have seen null used to say "yes this object exists and I am aware of it, but I am explicitly not storing anything against it (I may against other objects)" rather than "I don't know that this object exists".

I'd be happy to not support this in JS though, and keep it as-is. The only danger I can see is if users think setting values to null clears the cache entry: it will _look_ like this is the case, however it will retain a reference to the key, so would prevent garbage collection of the key if the script is unloaded. Again, probably enough of an edge case that we can ignore it for now.

florian-h05:
Okay, then I will do nothing.
I think that it is clear how to remove key-value from cache as there is `cache.remove(key)`.

Issue #88, opened 2022-02-04
Addition of timeUtils


rkoshak:
This issue is for the discussion of whether or not it makes sense to add my timeUtils library to openhab-js. I can see arguments either way but I've found this one of my libraries to be so incredibly useful I think it would be of great help to all OH users.

At a high level it has the following functions:

Function | Purpose | Dependencies
-|-|-
`parseDuration` | Parses a string like "2h5s" (two hours five seconds) into a `time.Duration`. | 
`durationToDateTime` | Returns a `time.Duration` or String added to `now`. | `parseDuration`
`isISO8601` | Returns true if the passed in string matches ISO8601 format. | 
`is24Hr` | Returns true if the passed in string matches 24 hour time format (HH:MM only). |
`is12Hr` | Returns true if the passed in string matches 12 hour time format (hh:MM aa only). |
`toToday` | Takes the passed in "when", converts it to a ZonedDateTime using `toDateTime`, and moves the date to today, taking into account DST. | `toDateTime`
`toTomorrow` | Like `toToday` but changes the date to tomorrow's date. | `toDateTime`
`toYesterday` | Like `toToday` but changes the date to yesterday's date. | `toDateTime`
`betweenTimes` | Returns true if `now` or the passed in "when" is between a start or end time, handling the case where the start and end times define a period that spans midnight. | `toDateTime`, `toTomorrow`, `toYesterday`
`toDateTime` | Converts almost anything you can think of to a `time.ZonedDateTime`. See below. | `is24Hr`, `is12Hr`, `durationToDateTime`, `parseDuration`

The types handled by `toDateTime`

Type | Returns
-|-
`time.ZonedDateTime` | returns when unmodified
`java.time.ZonedDateTime` | converts to `time.ZonedDateTime` (because of parsing issue (see #54) it uses the SYSTEM timezone)
JavaScript `Date` | converts to `time.ZonedDateTime`
String | If ISO-8601 format it will eventually parse it (see #54). If 24Hr time string, creates a `time.ZonedDateTime` with that time and today's date. If 12 Hr time string, creates a `time.ZonedDateTime` with that time and today's date, if duration string adds the resultant duration to `now`
`number` or `bigint` | Adds the value as milliseconds to now.
`DateTimeType` | extracts the Java ZonedDateTime and converts it to `time.ZonedDateTime`
`PercentType` | Adds the value to `now` as seconds.
`QuantityType` | If the units  are time units, adds the time to `now`.
[...]

PR #91, opened 2022-02-06
[WIP] Actions: Add timer manager

florian-h05:
## Idea

I have implemented a timer manager that allows the simple creation and modification of timers from UI and file-based rules.
It is utilizing the `sharedcache` to persist the access to the timer managment between reloads of scripts.

## Credits & Thanks

_Huge credits go to @rkoshak for his idea/implementation on [openhab-rules-tools](https://github.com/rkoshak/openhab-rules-tools/blob/main/timerMgr.js)._

## Motivation

I know that this capability is beyond the translation from Java to JS and reverse but I think timers are a really important thing for many rules and timer manager in the official library could help most users in bookkeeping and managing timers from rules.

For the advanced users out there, the access to the timers is still possible via `actions.ScriptExecution`..., therefore **this PR does not remove functionalities**.

## Notes

Please note that this PR depends on #90.
As fas as the improvements discussed in #88 are done, the `timerMgr` should utilize the `toDateTime` method (in line 37; have a look at TODO comment) to parse the time parameter.

***
@rkoshak @jpg0 @digitaldan Please let my know what you think of this addition, when it's okay I will contribute the docs.

### DIFF
--- a/actions/actions.js
+++ b/actions/actions.js
@@ -1,24 +1,6 @@
-/**
- * Actions namespace.
- * This namespace provides access to openHAB actions. All available actions can be accessed as direct properties of this
- * object (via their simple class name).
- * 
- * @example <caption>Sends a broadcast notification</caption>
- * let { actions } = require('openhab');
- * actions.NotificationAction.sendBroadcastNotification("Hello World!")
- * 
- * @example <caption>Sends a PushSafer notification</caption>
- * let { actions } = require('openhab');
- *  actions.Pushsafer.pushsafer("<your pushsafer api key>", "<message>", "<message title>", "", "", "", "")
- * 
- * @namespace actions
- */
-
-
-
-const osgi = require('./osgi');
+const osgi = require('openhab/osgi');
 const { actions } = require('@runtime/Defaults');
-const log = require('./log')('actions');
+const log = require('openhab/log')('actions');
 
 const Things = Java.type('org.openhab.core.model.script.actions.Things');
 const actionServices = osgi.findServices("org.openhab.core.model.script.engine.action.ActionService", null) || [];
--- a/actions/index.js
+++ b/actions/index.js
@@ -15,5 +15,6 @@
  */
 
 module.exports = {
-    ...require('./actions')
+    ...require('./actions'),
+    get timerMgr () { return require('./timerMgr'); }
 };
--- a/actions/index.js
+++ b/actions/index.js
@@ -1,0 +1,19 @@
+/**
+ * Actions namespace.
+ * This namespace provides access to openHAB actions. All available actions can be accessed as direct properties of this
+ * object (via their simple class name).
+ * 
+ * @example <caption>Sends a broadcast notification</caption>
+ * let { actions } = require('openhab');
+ * actions.NotificationAction.sendBroadcastNotification("Hello World!")
+ * 
+ * @example <caption>Sends a PushSafer notification</caption>
+ * let { actions } = require('openhab');
+ *  actions.Pushsafer.pushsafer("<your pushsafer api key>", "<message>", "<message title>", "", "", "", "")
+ * 
+ * @namespace actions
+ */
+
+module.exports = {
+    ...require('./actions')
+};
--- a/actions/package.json
+++ b/actions/package.json
@@ -1,0 +1,5 @@
+{
+    "name": "actions",
+    "version": "0.0.1",
+    "main": "index.js"
+}
--- a/actions/timerMgr.js
+++ b/actions/timerMgr.js
@@ -1,0 +1,95 @@
+const { cache, actions, log } = require('openhab');
+const logger = log('actions.timerMgr');
+
+/**
+ * Timer Manager namespace.
+ * This namespace allows creation and management of timers using the timer manager.
+ *
+ * Logger identifier is <caption>org.openhab.automation.script.actions.timermgr</caption>.
+ *
+ * @namespace actions.timerMgr
+ */
+
+/**
+ * Create a timer with the timerManager.
+ *
+ * If there is no timer with the given id, create one to expire at when and call func.
+ * If there already is a timer associated with id, then:
+ * If reschedule is true, reschedule the timer at when, else cancel the timer.
+ * If flappingFunc is provided, call flappingFunc.
+ *
+ * @memberOf actions.timerMgr
+ * @param {String} id the "name" of the timer
+ * @param {*} when any representation of time of duration
+ * @param {function} [func] function to call when the timer expires
+ * @param {boolean} [reschedule=false] optional flag, when present and true rescheudle the timer if it already exists
+ * @param {function} [flappingFunc] function to call when the timer already exists
+ *
+ * @example
+ * const now = require('openhab').time.ZonedDateTime.now();
+ * createTimer('my-timer', now.plusSeconds(10), () => { console.info('Success'); }, false, () => { console.warn('Already exists.'); });
+ */
+const createTimer = (id, when, func, reschedule, flappingFunc) => {
+  // Check for params.
+  if (!id) throw Error('Parameter id is required!');
+  if (!when) throw Error('Parameter when is required.');
+
+  const timeout = when; // TODO: needs toDateTime from rkoshak's timeUtils
+
+  const exists = cache.exists(id);
+  if (exists) {
+    // Timer already exists.
+    logger.debug('Timer [{}] already exists.', id);
+    const timer = cache.get(id);
+    if (reschedule) {
+      logger.debug('Rescheduling timer [{}].', id);
+      timer.reschedule(timeout);
+    } else {
+      logger.debug('Cancelling timer [{}].', id);
+      timer.cancel();
+      cache.remove(id);
+    }
+    if (flappingFunc) flappingFunc();
+  } else if (!exists) {
+    // Timer does not exist.
+    logger.debug('Creating timer [{}].', id);
+    const timer = actions.ScriptExecution.createTimer(timeout, () => {
+      // Debugging & Clean-Up.
+      logger.debug('Timer [{}] expired.', id);
+      cache.remove(id);
+      // Execute passed function.
+      if (func) func();
+    });
+    cache.put(id, timer);
+  }
+};
+
+/**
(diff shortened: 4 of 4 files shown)


jpg0:
I have a question regarding `setTimeout` / `setInterval` etc. These are the native JS equivalent to openHAB timers. #3 added polyfills so that indeed if you attempt to use these JS functions in scripts, then you do actually get openHAB timers under the covers.

My expectation is that for simple tasks, these should be sufficient. However for more complex management, they may not (at least on their own). This PR seems to include a little functionality over and above these functions, namely the storage of the timer in the cache, plus some associated things like `flappingFunc` related to this.

My questions are:
- If this additional functionality were built on the JS natives rather that timers directory (which of course use Timers anyway), what would it look like? Would it be simpler?
- If we do it this way (with native JS), how much code does it save the script writer, versus them using the JS natives themselves?

Typical code would look something like:
```
//to add
cache.put("myTimer", setTimeout(doStuff, 3000));

//to cancel
cache.remove("myTimer").clearTimeout();
```

I guess I'm wondering if there's enough benefit versus documenting a pattern using JS natives.

florian-h05:
> If this additional functionality were built on the JS natives rather that timers directory (which of course use Timers anyway), what would it look like? Would it be simpler?

You mean implementing the functionality of timer manager using `setTimeout` and `clearTimeout`?

The only disadvantage I see is, that the time parameter would have to be in milliseconds where this timer manager will support many different notations (as far as the `toDateTime` is present), e.g. "1h15
m". But it is a big disadvantage in my opinion. 
Except from that, I can still use the Java methods (reschedule, cancel) to manage the timer returned by setTimeout. 

> If we do it this way (with native JS), how much code does it save the script writer, versus them using the JS natives themselves?

You are talking about is this timer manager stuff needed/saves it much code?
So the rescheduling by just passing a flag and the flapping function are something that does not really require much code, but in a rule it would make a difference if I could just call:
'''js
timerMgr.check("test", "5m", doStuff, true, flappingFunc)
'''
Otherwise I had to write down the code for reschedule and flappingFunc over and over again as I am using timers in many rules.

I noticed that it would be better when the check() wouldn't remove the timer when reschedule flag is not set to true. 
An example for this: You have rule that schedules a timer to send an alarm when your door is open and temperature drops under a given level. When temp rises over that level and drops again, you would call check() again and it would either reschedule the alarm or cancel it. It would be better to reschedule or let the alarm how it is. To cancel a timer, you should explicitly call cancel().

digitaldan:
My 2 cents:  I think something like this should be provided outside of this library.  I understand the problem, but my thoughts are we should keep openhab-js as a thin layer on top of the existing openHAB functionality.  If openHAB  core had a concept of this, then i think it would make sense to expose, but since it does not, this seems better suited for community provided scripts.  

I think a good compromise would be to address pain points in the library that would assist such functionality.   For instance, i agree using milliseconds to schedule timers is not ideal for our use case that so often involves dates, so having a helper to convert a future ZoneDateTime to millis from now, or even have the `setTimeout` function take a ZoneDateTime as well as a number would seem reasonable.   If there are other helpers (like the function in #90 ), then lets look at adding those as well.

florian-h05:
> I think something like this should be provided outside of this library. I understand the problem, but my thoughts are we should keep openhab-js as a thin layer on top of the existing openHAB functionality. If openHAB core had a concept of this, then i think it would make sense to expose, but since it does not, this seems better suited for community provided scripts.

Okay, I understand that thoughts. When we create a CONTRIBUTING.md (maybe taking that https://github.com/openhab/openhab-addons/blob/main/CONTRIBUTING.md and modifying it to our needs), that should go into them. 

> I think a good compromise would be to address pain points in the library that would assist such functionality.

Okay, as we obviously won't include the timer manager, for me that's a good compromise.

> [...] have the setTimeout function take a ZoneDateTime as well as a number would seem reasonable.

I would have to make changes in https://github.com/openhab/openhab-addons/blob/main/bundles/org.openhab.automation.jsscripting/src/main/resources/node_modules/%40jsscripting-globals.js, right?

I will have a look at having `setTimeout` take ZonedDateTime as well as a number (if the passed delay argument is integer I would always suppose it is milliseconds). 

**I am wondering if we should also create a function like `rescheduleTimeout` to reschedule.**

For development, I would define that functions in my test script as building the JS Scripting Add-On

jpg0:
@digitaldan completely agree, a much more direct statement of what I was getting at with my comments. I would add that we should try to be adhering to the unix philosophy of small, reusable, composable functions, whereas this example feels  bit to high-level and over-specific to a single use-case.

I do think that we should have a stance on things like accepting ZonedDateTime in `setTimeout` as this diverges from the JS standard. Technically it should mean that standard JS code can be ported to openHAB fine (but would make porting away from openHAB harder). It may however break tooling, especially code completion/syntax highlighting. My personal view is that it feels wrong (and may be a slippery slope) and I would prefer that we have a simple way to convert a ZonedDateTime to millis-from-now, and keep the standard JS object as close to the standard as possible.

(The other option that I alluded to would be to provide an openHAB-specific set of timer functions that wrapped and augmented the JS native ones, where we could allow automatic conversion plus rescheduling etc. But that would certainly be less reusable.)

florian-h05:
> My personal view is that it feels wrong (and may be a slippery slope) and I would prefer that we have a simple way to convert a ZonedDateTime to millis-from-now, and keep the standard JS object as close to the standard as possible.

Yes, it feels wrong to implement standard JS with non standard function.

> The other option that I alluded to would be to provide an openHAB-specific set of timer functions that wrapped and augmented the JS native ones, where we could allow automatic conversion plus rescheduling etc. But that would certainly be less reusable.

In my view that is a comparable situation as with the timer manager PR now. 

To sum it up: 
- **We decided to go with a ZonedDateTime to millis-from-now function?**
- **What do you thing about implementing something like `rescheduleTimeout` as a global function (although standard JS has nothing comparable)?**

jpg0:
> In my view that is a comparable situation as with the timer manager PR now.

Functionally it is, it's just likely to result in simpler code, plus keeps the interaction with Java Timer objects only in the polyfill.

> What do you thing about implementing something like rescheduleTimeout as a global function (although standard JS has nothing comparable)?

I'm not sure how this would work. What do you pass into this function?

What do we lose if we don't support this (and instead require cancellation + recreation)?

rkoshak:
I'm a little late to the party but I agree that a function like this is probably best handled by a community library (and in fact already is in openhab_rules_tools which can now be installed via npm). 

But there seems to be a lot of overlap between what's being discussed here and what is being discussed on #88. And I'm not sure if the there is some contradictory suggestions being made. I've started down the path of monkey patching a `toDateTime()` method to easily convert stuff to a Joda DateTime (pending some experiments to make sure that it works in all cases, such as when the user decides to `Java.type` the class themselves). 

But here I get the impression that we don't want to do that because it could make migrating code to/from openHAB more difficult. I haven't done much yet on this so it's all good either way. But I do want to get it right. 1. Should I keep is completely separate (e.g. some functions in utils.js), 2. monkey patch time.ZonedDateTime with a new `from` function, or 3. do I continue down the path of adding a `toDateTime()` to String, and java.time.ZonedDateTime, etc.?

If 1 I can have a PR submitted in an hour or so. If 2 or 3 it'll take a bit more time. Not that the time matters, I just want to get it right.

florian-h05:
> I'm not sure how this would work. What do you pass into this function?

This function would take two arguments: the timer returned by `setTimeout` and the new delay. 

> What do we lose if we don't support this (and instead require cancellation + recreation)?

We loose nothing, I think it would be possible to save the timer returned by `setTimeout` (e.g. in cache) and then run the Java method `reschedule()` in it.
If we would choose this way, I would definitely mention this in docs.

jpg0:
When I asked what we would loose, what I meant was whether it was reasonable to ask users to call `clearTimeout` then `setTimeout` to recreate the timer rather than reschedule it. I ask this because JS developers seem to be fine with only these two things, and it means no more work here.

If we do choose to support rescheduling, I would suggest a slight tweak on your proposal. The `setTimeout` function [returns a timeoutID, which is an integer](https://developer.mozilla.org/en-US/docs/Web/API/setTimeout). This timeoutID can be then used to cancel the execution, via `clearTimeout(timeoutID)`. Let's keep these functions as-is, to stick with the JS standard. However, we could certainly provide a `reschedule` or `updateTimeout` or something, that accepts a timeoutID. (I would put this into a timer package or something.) This would allow us to both adhere to the JS standard, but provide this extra functionality.

florian-h05:
> The setTimeout function [returns a timeoutID, which is an integer](https://developer.mozilla.org/en-US/docs/Web/API/setTimeout).

Looking at https://github.com/openhab/openhab-addons/blob/main/bundles/org.openhab.automation.jsscripting/src/main/resources/node_modules/%40jsscripting-globals.js#L176:
- The polyfilled `setTimeout` of the JS Scripting Add-On differs from the native JS. `setTimeout` of the JS Scripting Add-On returns an openHAB Timer instead of a timeoutID.
- `clearTimeout` differs in the parameter: it takes an openHAB Timer instead of the timeoutID.
- **Functionally, that does not make any difference to native JS, but it allows us to perform the methods of the [Timer Interface](https://www.openhab.org/javadoc/latest/org/openhab/core/model/script/actions/timer) on the returned timer.**

What I want to say, is that it is possible to directly call `reschedule(time)` on the timer returned by `setTimout`:
```js
var timer = setTimeout(func, 10000);
timer.reschedule(time.ZonedDateTime.now().plusNanos(1000 * 1000000));
```
But it does not look „right“.

> However, we could certainly provide a reschedule or updateTimeout or something, that accepts a timeoutID. (I would put this into a timer package or something.) This would allow us to both adhere to the JS standard, but provide this extra functionality.

That would be great, as it would allow to reschedule the timer by using milliseconds as argument to be close to native JS and to what `setTimeout` accepts.

## To summarize:
- I will close this PR and create a new one, that:
- Adds a `updateTimeout` function to the `time` namespace of openhab-js that accepts two parameters: the openHAB timer (that was returned by `setTimeout`) and the new delay in milliseconds.
- Also add a method to namespace `utils` or `time` that parses from ZonedDateTime to milliseconds from now.

florian-h05:
New PR is #93.

PR #92, opened 2022-02-06
Re-Arrange .gitignore & Add .npmignore

florian-h05:
Use .npmignore to reduce package size.

Testing the .npmignore (with `npm pack`) shows a decrease in size:

- package size: 489.8 kB -> 57.5 kB
- unpacked size: 836.0 kB -> 239.0 kB

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
-node_modules
-out
+.vscode
 dist
 docs
-.vscode
+node_modules
+openhab-*.tgz
+out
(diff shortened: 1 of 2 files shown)


digitaldan:
Since npm will use a .gitignore file if no .npmignore file is present, is there a benefit of maintaining 2 difference files?

florian-h05:
Yes, the benefit is that we can npmignore folders and files we can not ignore using gitignore.

For example the images and the test folder, we need them for docs and for development, but we do not need them in our node package.

digitaldan:
Gotcha makes sense, thanks!

PR #93, opened 2022-02-08
Time: Add updateTimeout & ZDTtoMillisFromNow

florian-h05:
Replaces #91.

This PR adds two methods to the `time` namespace:
- `updateTimeout` can reschedule a timer created by `setTimeout`.
- `ZDTtoMillisFromNow` parses the duration from now to the given ZDT in milliseconds.

Additionally, it updates the docs according to the changes.
These changes were announced at #91.

@jpg0 @digitaldan 
Happy with this solution?


jpg0:
I'd certainly like the opinion of @digitaldan here. Specifically on the overall approach to how we handle openHAB dependency objects - such as ZonedDateTime, when we need additional functionality related to another (3rd party object), such as conversions. The options as I see them are:
- extend the openHAB object (for example `ZonedDateTime.millisFromNow()`)
- extend the other side of the conversion, mapping into openHAB objects (**not** a great example is `1000.toZonedDateTime()` because the object is an integer in this case, but you get the idea)
- provide utility mapping functions (like in this PR)

This completely also relates to the work that @rkoshak has been doing and his points around conflicting approaches.

I originally said that I preferred to avoid the last option as it leads to a grab-bag full of utility functions, and composition  of logic is hard (nesting functions rather than chaining them), however too much patching will also lead to problems.

digitaldan:
>, and composition of logic is hard (nesting functions rather than chaining them), however too much patching will also lead to problems.

Thats definitely the issue here for sure.  As a user, i would far prefer to use `myzdt.millisFromNow()` then use  `time. ZDTtoMillisFromNow(myzdt)` , but i get the code smell of patching a common library like JS-Joda. 

That being said, @rkoshak 's PR #95 does exactly this, so if we do go down that path  , then i think it makes sense to do that here as well.  I have to admit i find that extending ZoneDateTime seems  to feel way more user friendly.  I think i could justify this if we had a section of the README which specifically calls out our extensions and that we try to be very disciplined about what we add (and not go crazy) .  I'm not worried about upstream JS-Joda changes breaking this, that library is pretty much set in stone and will only be getting security and bug updates IMHO.  

The timer reschedule logic does seem out of place here since we don't really deal with native JS timers in this package.   While the setTimeout/SetInterval standard does specify a integer to refer to timer instances, i did opt to just use the underlying timer object instead as A) i didn't want to try and keep some artificial map of ints -> timers and B) i didn't think anyone was really going to notice or care :-)   So yes a user can easily reschedule a timer by doing `timer.reschedule(...);` on something returned from setTimeout/setInterval.  Maybe it would be better to be transparent about this in the doc and not create another method that obfuscates this ?

digitaldan:
Oh and sorry for the long delay, its been a very busy new year at work and i have had little time to dedicate to openHAB !

florian-h05:
> I think i could justify this if we had a section of the README which specifically calls out our extensions and that we try to be very disciplined about what we add (and not go crazy) .

Okay, then I will  work on a `ZonedDateTime.millisFromNow()` function instead of adding the `ZDTtoMillisFromNow(zdt)`.

> So yes a user can easily reschedule a timer by doing timer.reschedule(...); on something returned from setTimeout/setInterval. Maybe it would be better to be transparent about this in the doc and not create another method that obfuscates this ?

Hmm, my idea behind creating that function was to support the rescheduling with milliseconds to follow the style of setTimeout. 
But you are right with your thoughts about transpareny and it might be confusing to have something like `updateTimeout` that looks like standard JS but is not standard JS.
From my personal opinion, we can also go with updating the docs to show how to reschedule.

> Oh and sorry for the long delay, its been a very busy new year at work and i have had little time to dedicate to openHAB !

No problem :-), we all have other commitments than openHAB.

florian-h05:
@digitaldan 
`millisFromNow()` is monkey-patched on ZonedDateTime and a section about the monkey patches is added to the README.
The `updateTimeout` method is removed, instead a section about the returned openHAB Timer is added to README.

I noticed that JSDoc doesn't generate docs for monkey-patched functions. That might be interesting for @rkoshak.

rkoshak:
I didn't think it would generate docs but think it's still worth commenting or code as if it did.

Once I get a review on my PR I plan on submitting another to the add-on readme to discuss these new features. I didn't want to start that until my approach was approved since it works change significantly if I just monkey patch ZonedDateTime or provide a utility function instead.

Also note that I don't really have a dog in this fight. Whether we monkey patch or not it's not something I feel strongly about either way. So even though I've submitted that PR don't treat that longer I've staked out a position. I'm equally as happy with all three approaches. I just want to be consistent across the library.

time.js
```diff
@@ -1,12 +1,29 @@
...
+
+/**
+ * Parses a ZonedDateTime to milliseconds from now until the ZonedDateTime.
+ * This is a monkey-patched function on JS-Joda's ZonedDateTime.
+ *
+ * @returns {number} duration from now to the ZonedDateTime in milliseconds
+ */
+time.ZonedDateTime.prototype.millisFromNow = function () {
+  const duration = time.Duration.between(time.ZonedDateTime.now(), this);
+  return duration.toMillis();
```
digitaldan, line 26:
```suggestion
  return time.Duration.between(time.ZonedDateTime.now(), this).toMillis();
```

time.js
```diff
@@ -1,12 +1,29 @@
...
 const targetParse = time.ZonedDateTime.prototype.parse;
 time.ZonedDateTime.prototype.parse = function (text, formatter = rfcFormatter) {
-    return targetParse(text, formatter);
-}
+  return targetParse(text, formatter);
+};
+
+/**
+ * Parses a ZonedDateTime to milliseconds from now until the ZonedDateTime.
+ * This is a monkey-patched function on JS-Joda's ZonedDateTime.
```
digitaldan, line 20: I don't think we need to add the monkey-patched comment here .

florian-h05:
@digitaldan 
I made your requested changes.

README.md
```diff
@@ -453,6 +465,9 @@ openHAB internally makes extensive use of the `java.time` package.
 openHAB-JS exports the excellent [JS-Joda](#https://js-joda.github.io/js-joda/) library via the `time` namespace, which is a native Javascript port of the same API standard used in Java for `java.time`.
 Anywhere that a native Java `ZonedDateTime` or `Duration` is required, the runtime will automatically convert a JS-Joda `ZonedDateTime` or `Duration` to its Java counterpart.
 
+The openHAB-provided [JS-Joda](#https://js-joda.github.io/js-joda/) library is extended with convenient functions relevant to openHAB usage.
+Those extensions are listed in [openHAB-specific extensions for JS-Joda](#openhab-specific-extensions-for-js-joda).
```
digitaldan, line 469:
```suggestion
#### openHAB-JS extensions to JS-Joda
The  exported [JS-Joda](#https://js-joda.github.io/js-joda/) library is also extended with convenient functions relevant to openHAB usage. 

* `ZonedDateTime`
  * `millisFromNow()` => Returns the milliseconds from now until the ZonedDateTime representation.

```

README.md
```diff
@@ -466,6 +481,10 @@ console.log("averageSince", item.history.averageSince(yesterday));
 actions.Exec.executeCommandLine(time.Duration.ofSeconds(20), 'echo', 'Hello World!');
 ```
 
+#### openHAB-specific extensions for JS-Joda
+* `ZonedDateTime`
+  * `millisFromNow()` => Returns the milliseconds from now until the ZonedDateTime representation.
+
```
digitaldan, line 487:
```suggestion
```

README.md
```diff
@@ -453,6 +465,9 @@ openHAB internally makes extensive use of the `java.time` package.
 openHAB-JS exports the excellent [JS-Joda](#https://js-joda.github.io/js-joda/) library via the `time` namespace, which is a native Javascript port of the same API standard used in Java for `java.time`.
 Anywhere that a native Java `ZonedDateTime` or `Duration` is required, the runtime will automatically convert a JS-Joda `ZonedDateTime` or `Duration` to its Java counterpart.
 
+The openHAB-provided [JS-Joda](#https://js-joda.github.io/js-joda/) library is extended with convenient functions relevant to openHAB usage.
+Those extensions are listed in [openHAB-specific extensions for JS-Joda](#openhab-specific-extensions-for-js-joda).
+
 Examples:
```
digitaldan, line 471: Should we add an example of millisFromNow() here ?

README.md
```diff
@@ -154,7 +154,19 @@ var timeoutID = setTimeout(function[, delay]);
 
 The global `clearTimeout()` method cancels a timeout previously established by calling `setTimeout()`.
 
-see https://developer.mozilla.org/en-US/docs/Web/API/setTimeout for more information about setTimeout.
+See https://developer.mozilla.org/en-US/docs/Web/API/setTimeout for more information about setTimeout.
+
+openHAB does not return the integer timeout id as standard JS does, instead it returns an openHAB Timer.
```
digitaldan, line 159: I am thinking we should have a `Timers` section with 3 subsections, so move  `SetTimeout` and `SetInterval` here and add `openHAB Timers` as well.     We then can mention in  `SetTimeout` and `SetInterval` that they return a native openHAB timer instance and to see `openHAB Timers` for more info.  In this section we can talk about rescheduling and other functions found in  https://www.openhab.org/javadoc/latest/org/openhab/core/model/script/actions/timer and give specific examples.

WDYT?

digitaldan (changes requested):

README.md
florian-h05: We could add it to the first block of examples.

README.md
florian-h05: Yeah, that should be the best option as it is well structured.
I will have a look at that.

florian-h05:
@digitaldan 
Ready to re-review.

PR #99, opened 2022-02-17
Implemented a toggle method on Item

rkoshak:
Signed-off-by: Richard Koshak <rlkoshak@gmail.com>

# [Item] Added a Toggle Method to Item

# Description

Added a `toggle` method to Item which will toggle the state of the Item based on its current state. Any Item that supports OnOffType command or update will receive the opposite of its current `getStateAs(OnOffType)` as an update or command. Any Item that supports OpenClosedType updates will receive the opposite of its current `getStateOs(OpenClosedType)`. In all other cases an error is thrown (including attempting to toggle a Contact with a command). There is an argument `update` which, when provided with `true` will cause the new toggled state to be posted as an update. The default is `false` causing the new toggled state to be sent as a command.

There are many situations where one needs to toggle a binary type Item and this saves the user from needing to repeat that sort of logic over and over, similar to the already present `sendCommandIfDifferent()`.

# Testing

I tested most of the different permutations including:

- toggle on an Item whose state is UnDefType: generates error
- toggle on Switch without `update` argument: Item was commanded
- toggle on Switch with `update=false`: Item was commanded
- toggle on Switch with `update=true`: Item was updated
- toggle on Contact with `update=false`: generates error
- toggle on Contact with `update=true`: Item was updated

### DIFF
--- a/README.md
+++ b/README.md
@@ -146,7 +146,7 @@ see https://developer.mozilla.org/en-US/docs/Web/API/console for more informatio
 
 ### SetTimeout
 
-The global `setTimeout()` method sets a timer which executes a function or specified piece of code once the timer expires.
+The global setTimeout() method sets a timer which executes a function or specified piece of code once the timer expires.
 ```javascript
 var timeoutID = setTimeout(function[, delay, arg1, arg2, ...]);
 var timeoutID = setTimeout(function[, delay]);
@@ -154,7 +154,7 @@ var timeoutID = setTimeout(function[, delay]);
 
 The global `clearTimeout()` method cancels a timeout previously established by calling `setTimeout()`.
 
-See https://developer.mozilla.org/en-US/docs/Web/API/setTimeout for more information about setTimeout.
+see https://developer.mozilla.org/en-US/docs/Web/API/setTimeout for more information about setTimeout.
 
 ### SetInterval
 
@@ -304,7 +304,7 @@ See [openhab-js : actions.Audio](https://openhab.github.io/openhab-js/actions.ht
 
 #### BusEvent
 
-See [openhab-js : actions.BusEvent](https://openhab.github.io/openhab-js/actions.html#.BusEvent) for complete documentation.
+See [openhab-js : actions.BusEvent](https://openhab.github.io/openhab-js/actions.html#.BusEvent) for complete documentation
 
 #### Ephemeris Actions
 
--- a/README.md
+++ b/README.md
@@ -27,8 +27,7 @@ binding](https://www.openhab.org/addons/automation/jsscripting/)
 - [Scripting Basics](#scripting-basics)
   - [Require](#require)
   - [Console](#console)
-  - [SetTimeout](#settimeout)
-  - [SetInterval](#setinterval)
+  - [Timers](#timers)
   - [ScriptLoaded](#scriptloaded)
   - [ScriptUnLoaded](#scriptunloaded)
   - [Paths](#paths)
@@ -144,9 +143,11 @@ The string representations of each of these objects are appended together in the
 
 see https://developer.mozilla.org/en-US/docs/Web/API/console for more information about console logging.
 
-### SetTimeout
+### Timers
 
-The global setTimeout() method sets a timer which executes a function or specified piece of code once the timer expires.
+#### SetTimeout
+
+The global `setTimeout()` method sets a timer which executes a function or specified piece of code once the timer expires.
 ```javascript
 var timeoutID = setTimeout(function[, delay, arg1, arg2, ...]);
 var timeoutID = setTimeout(function[, delay]);
@@ -154,9 +155,11 @@ var timeoutID = setTimeout(function[, delay]);
 
 The global `clearTimeout()` method cancels a timeout previously established by calling `setTimeout()`.
 
-see https://developer.mozilla.org/en-US/docs/Web/API/setTimeout for more information about setTimeout.
+See https://developer.mozilla.org/en-US/docs/Web/API/setTimeout for more information about setTimeout.
 
-### SetInterval
+openHAB does not return the integer timeoutID as standard JS does, instead it returns an instance of [openHAB Timer](#openhab-timer).
+
+#### SetInterval
 
 The setInterval() method repeatedly calls a function or executes a code snippet, with a fixed time delay between each call.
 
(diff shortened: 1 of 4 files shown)


rkoshak:
Would it make sense to update/command the Item if its current state is `NULL` or `UNDEF` instead of throwing an error? It occurs to me that with the current error it might confuse users into thinking that the problem is the Item type and not that it's current state is the problem.

items/managed.js
```diff
@@ -182,6 +182,28 @@ class Item {
...
+        var oldState = this.rawItem.getStateAs(OnOffType);
+        if(oldState) {
+            newState = (oldState.toString() == 'OFF') ? 'ON' : 'OFF';
+        }
+        else if(update && this.rawItem.getStateAs(OpenClosedType)) {
+            newState = (this.rawItem.getStateAs(OpenClosedType).toString() == 'CLOSED' ? 'OPEN' : 'CLOSED');
+        }
+        if(newState && update) this.postUpdate(newState);
+        else if(newState && !update) this.sendCommand(newState);
+        else throw Error('Items of type ' + this.rawItem.getType() + ' cannot be toggled via ' + ((update) ? 'an update' : 'a command'));
```
digitaldan, line 204:
```suggestion
        if (newState) { 
          update ? this.postUpdate(newState) : this.sendCommand(newState);
        } else  { 
          throw Error('Items of type ' + this.rawItem.getType() + ' cannot be toggled via ' + ((update) ? 'an update' : 'a command'));
       }
```

digitaldan, line 204:
```suggestion
        if (newState) { 
          update ? this.postUpdate(newState) : this.sendCommand(newState);
        } else  { 
          throw Error('Items of type ' + this.rawItem.getType() + ' cannot be toggled via ' + ((update) ? 'an update' : 'a command'));
       }
```

items/managed.js
```diff
@@ -182,6 +182,28 @@ class Item {
...
+     * Toggles the state of the Item. If the Item is undefined or the Item type
+     * does not support OnOffType commands/updates or OpenClosedType updates
+     * an error is thrown. To toggle Contacts, update must be true or an error
+     * is thrown (Contacts cannot be commanded).
+     * @param {Boolean} [update=false] when true the toggle will be sent as an update instead of a command
+     * @throws error if the Item cannot be toggled
+     */
+    toggle(update=false) {
+        var newState = null;
+        var oldState = this.rawItem.getStateAs(OnOffType);
```
digitaldan, line 195:
```suggestion
        let newState = null;
        let oldState = this.rawItem.getStateAs(OnOffType);
```

minor nit :-)

items/managed.js
```diff
@@ -182,6 +182,28 @@ class Item {
...
+        if(oldState) {
+            newState = (oldState.toString() == 'OFF') ? 'ON' : 'OFF';
```
digitaldan, line 197:
```suggestion
            newState = oldState.toString() == 'OFF' ? 'ON' : 'OFF';
```

items/managed.js
```diff
@@ -182,6 +182,28 @@ class Item {
...
+        }
+        else if(update && this.rawItem.getStateAs(OpenClosedType)) {
+            newState = (this.rawItem.getStateAs(OpenClosedType).toString() == 'CLOSED' ? 'OPEN' : 'CLOSED');
```
digitaldan, line 200:
```suggestion
            newState = this.rawItem.getStateAs(OpenClosedType).toString() == 'CLOSED' ? 'OPEN' : 'CLOSED';
```

digitaldan (changes requested):
I made a couple comments, one thing that seems odd to me about this is that we have one method which handles both a command and an update, which is kinda awkward to explain given some types support update and commands and some don't.  I get toggling switch like items and sending a command, rolling in updates and contacts seems to confuse this.  Maybe we should stick to what we support in rules as well ? 

 https://github.com/openhab/openhab-js/blob/d1b3cd5f87e967b59d8c612efc38bc83b7b3eb81/rules/operation-builder.js#L405

we could probably move this onto an item and reuse/reduce the amount of code needed.

items/managed.js
rkoshak: Will do

items/managed.js
rkoshak: Will do

items/managed.js
rkoshak: Will do

rkoshak:
I think I may have accidentally rebaselined this PR. I thought I was on main on my fork. 

That wasn't my intention. I'm not sure how to proceed from here.

Clearly git and I do not get along. :-(

Anyway, I've addressed all the comments from above. I added a toggleCommand and toggleUpdate to Item and changed doToggle to call toggleCommand so it's all centralized.

rules/operation-builder.js
```diff
@@ -404,21 +404,7 @@ class ToggleOperation extends OperationConfig {
      */
     doToggle() {
         let item = items.getItem(this.itemName);
```
digitaldan, line 406: Since we only use the reference once, might be cleaner to do
```suggestion
        items.getItem(this.itemName).toggleCommand();
```

time.js
```diff
@@ -1,12 +1,28 @@
...
+};
+
+/**
+ * Parses a ZonedDateTime to milliseconds from now until the ZonedDateTime.
+ *
+ * @memberof time
+ * @returns {number} duration from now to the ZonedDateTime in milliseconds
+ */
+time.ZonedDateTime.prototype.millisFromNow = function () {
+  return time.Duration.between(time.ZonedDateTime.now(), this).toMillis();
```
digitaldan, line 25: Not sure if this should be in this PR?

README.md
```diff
@@ -144,19 +143,23 @@ The string representations of each of these objects are appended together in the
 
 see https://developer.mozilla.org/en-US/docs/Web/API/console for more information about console logging.
 
-### SetTimeout
+### Timers
```
digitaldan, line 146: I think this is from a another PR, as well as all the other time changes?

digitaldan (changes requested):
Thanks Richard, it looks like there are a lot of changes related to your PR around time, I also left a very small comment.  Can you  fix this so it just includes toggle functionality?  Might be easier to start fresh and cherry pick changes

time.js
rkoshak: Yes, I mucked things up and somehow merged upstream on this branch when I meant to on main instead. That's not actually my code. I'll try to get back to just my PR.

README.md
rkoshak: See above

rules/operation-builder.js
rkoshak: Done

README.md
rkoshak: Done

items/managed.js
```diff
@@ -181,6 +181,55 @@ class Item {
...
+            }
+        }
+    }
+
+    /**
+     * Sends a command to flip the Item's state (e.g. if it is 'ON' an 'OFF' 
+     * command is sent).
+     * @throws error if the Item is uninitialized or a type that cannot be toggled or commanded
+     */
+    toggleCommand() {
```
digitaldan, line 218:
```suggestion
    sendToggleCommand() {
```
wdyt?

items/managed.js
```diff
@@ -181,6 +181,55 @@ class Item {
...
+        }
+        this.sendCommand(this.#getToggleState());
+    }
+
+    /**
+     * Posts an update to flip the Item's state (e.g. if it is 'ON' an 'OFF'
+     * update is posted).
+     * @throws error if the Item is uninitialized or a type that cannot be toggled
+     */
+    toggleUpdate() {
```
digitaldan, line 230:
```suggestion
    postToggleUpdate() {
```
wdyt?  I'm thinking we should try and be consistant since we use `postUpdate` elsewhere

digitaldan (changes requested):
Thanks @rkoshak !  Looks pretty good to me, i only had 2 small comments about naming

items/managed.js
rkoshak: That works for me. I like consistency.

items/managed.js
rkoshak: Same as above. I'll get a PR submitted later today with these two changes.

digitaldan:
Hey @rkoshak any update on this?

rkoshak:
I've been trying to figure out how to service the unit tests using Mocha but haven't had a lot of time these last few weeks to do much of anything. I've but forgotten and am still working it. I can commit the charges without the tests to get this moving.

items/managed.js
rkoshak: Done

items/managed.js
rkoshak: Done

rkoshak:
I gave up on the uint testing. I might come back at a date with a new PR once I figure it out. Your requested changes have been made with this latest PR.

rules/operation-builder.js
```diff
@@ -403,22 +403,7 @@ class ToggleOperation extends OperationConfig {
...
-            }
-            case "ColorItem": {
-                let toSend = ('0' != item.rawState.getBrightness().toString()) ? 'OFF' : 'ON';
-                item.sendCommand(toSend);
-                break;
-            }
-            default:
-                throw Error(`Toggle not supported for items of type ${item.type}`);
-        }
+        items.getItem(this.itemName).toggleCommand();
```
florian-h05, line 406: Wouldn't this be `sendToggleCommand` instead of `toggleCommand`?

florian-h05 (commented):
LGTM overall, but please have a look at my comment!

digitaldan:
@rkoshak i'm ok without the testing , if you want to fix the one issue mentioned i'm happy to merge, thanks!

rules/operation-builder.js
rkoshak: Fixed

rkoshak:
Should be good to go now.

florian-h05 (approved):
LGTM, thank you @rkoshak.
I did no testing but it should work.

digitaldan:
lgtm thanks!

PR #106, opened 2022-04-01
Docs: Add the event object of file-based rules

florian-h05:
# Docs: Add the `event` object of file-based rules to the README

# Description

This PR adds the documentation for the `event` object in file-based rules to the README.

@digitaldan @jpg0
The `event` object of UI rules completely differs from the file-based one, when I dump the event of a UI rule I get the Java classes of the triggers and need to use the Java methods on them to get the `receivedCommand` and so on.
Can we improve that to have the same object as in file-based with properties??

# Testing

You may use something like that to get some information about the event object for a trigger:
```javascript
const { rules, triggers, utils } = require('openhab');

rules.JSRule({
  name: 'Event Object file-based Tests',
  triggers: triggers.TimeOfDayTrigger('18:23'),
  execute: (event) => {
    utils.dumpObject(event);
    const keys = Object.keys(event);
    for (const i in keys) console.info('event.' + keys[i] + ' is ' + event[keys[i]] + ' typeof ' + typeof event[keys[i]]);
  }
});
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,6 +41,7 @@ binding](https://www.openhab.org/addons/automation/jsscripting/)
 - [File Based Rules](#file-based-rules)
   - [JSRule](#jsrule)
   - [Rule Builder](#rule-builder)
+  - [Event Object](#event-object)
 
 ## Installation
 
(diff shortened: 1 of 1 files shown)


jpg0:
@florian-h05 I believe that this difference manifests elsewhere too - such as not being able to use `const` in UI based rules because script environments are reused. The right approach would be to reuse the same implementation across both file-based and UI-based scripts.

florian-h05:
> The right approach would be to reuse the same implementation across both file-based and UI-based scripts.

Yes, that would be great. Getting UI-based closer to file-based would be good as file-based is closer to native JS, e.g. with `let` and `const`.
But this is beyond my skills and knowledge.

digitaldan:
Thanks @florian-h05, the table of event properties should be very helpful.

Issue #102, opened 2022-03-04
Add utilities to improve QuantityType handling


rkoshak:
QuantityTypes are somewhat of a pain to work with. I think the current implementation in openhab-js leaves much to be desired.

My understanding is that currently the library goes out of it's way to throw away the units for QuantityTypes and pretty much forces the users who actually want to use units in their calculations to revert to messing with the rawState and importing and creating QuantityTypes themselves. 

This issue is to discuss and track ideas for ways to make it easier for end users to handle QuantityTypes.

Unfortunately, based on my research and experiments I've determined the following:

- JavaScript does not support operator overloading so we can't patch something in to make +, -, et. al. handle this.
- Try as I might, I could not monkey patch a Java class.

So we are looking at adding `add()`, `subtract()`, et. al. helper functions *somewhere*.

I've already done some experimentation using this approach and I think it could work. What I propose is adding all the operations as helper functions (to mirror the Java QuantityType I'd just create a compareTo instead of equals, lt, gt, et. al. functions). Each of these functions accept two arguments and follow these rules:

- if both are QT, make sure they are compatible units and if so, do the operation, throw an exception otherwise
- if one is any sort of number and the other is QT, convert the number to a QT of the same units and do the operation
- if both are any sort of number, do the operation as normal
- if one is a string representing a number with units (e.g. '65 °F'), convert that to a QT and then follow the rules above.

In experimenting I've basically already implemented everything but that last rule in a prototype and it all works as expected (though there are some gotchas with multiply and divide which I may need to think about, namely is it even reasonable to convert the non-QT operand to a QT?).

Before I go and create a new PR (given the trouble I've had already I want to wait until my outstanding PRs are merged before creating new ones) I want to know:

1. Does this sound like a good idea? It's probably going to be less useful to people who are writing rules in files or who are very well aware of the types of their Items but I think it's going to be critical for Blockly users when Blockly moves to JS Scripting.
2. Where to put it? utils.js? Item?

## Your Environment

* Version used: (e.g., openHAB and **JS Scripting add-on version**)
* Environment name and version (e.g. Chrome 76, Java 8, Node.js 12.9, ...):
* Operating System and version (desktop or mobile, Windows 10, Raspbian Buster, ...):

jpg0:
I'm posting now as I've only just been learning about QuantityTypes...

[...]

PR #110, opened 2022-04-24
Docs: document UI rules event and @runtime

ssalonen:
Open questions
- is `@runtime` available in file based rules? If not, it should be mentioned in docs

Not documented
- trigger rules in UI
- cron rules in UI
- ...


README.md
```diff
@@ -15,41 +15,68 @@ It provides convenient access to common openHAB functionality within rules inclu
...
 
 ### Default Installation
 
-- Install the openHAB [JavaScript binding](https://www.openhab.org/addons/automation/jsscripting/), a version of this
-library will be automatically installed and available to all javascript rules by default.
+By default, openHAB has fairly old javascript runtime (in MainUI `ECMA - 262 Edition 5.1` or `application/javascript`). In [community forum](https://community.openhab.org) many people also call this as "Nashorn engine".
 
+This library is not compatible with this old runtime.
+
+Instead, this library is meant to be used with ECMAScript 2021+ rules (also known as `ECMAScript 262 Edition 11` or `application/javascript;version=ECMAScript-2021`). These rules created using files ([File Based Rules](#file-based-rules)) or MainUI ([UI Based Rules](#ui-based-rules)).
```
florian-h05, line 77: I would write something like:

> Instead, this library is meant to be used with the ECMAScript 2021+ engine (also known as `ECMAScript 262 Edition 11` or `application/javascript;version=ECMAScript-2021`). Rules can be created using files ([File Based Rules](#file-based-rules)) or MainUI ([UI Based Rules](#ui-based-rules)).

README.md
```diff
@@ -108,6 +135,40 @@ console.log("Thing status",thingStatusInfo.getStatus());
 
 See [openhab-js](https://openhab.github.io/openhab-js) for a complete list of functionality
 
+### Event Object
+
+**NOTE**: Note that `event` object is different in UI based rules and File based rules! This section is only valid for UI based rules.
```
florian-h05, line 140: I would spell `File` in lowercase (`file`).
Maybe you could add a link to the documentation for the event in file-based rules.

README.md
```diff
@@ -108,6 +135,40 @@ console.log("Thing status",thingStatusInfo.getStatus());
...
 
+### Event Object
+
+**NOTE**: Note that `event` object is different in UI based rules and File based rules! This section is only valid for UI based rules.
+
+When you use "Item event" as trigger (i.e. "[item] received a command", "[item] was updated", "[item] changed"), there is additional context available for the action in a variable called `event`.
+
+This tables gives an overview over the `event` object for most common trigger types:
+
+| Property Name  | Type                                                                                                                 | Trigger Types                          | Description                                                                                                   |                        |     |
```
florian-h05, line 146: Can please remove that last empty column?

README.md
```diff
@@ -108,6 +135,40 @@ console.log("Thing status",thingStatusInfo.getStatus());
...
+| `itemCommand`  | sub-class of [org.openhab.core.types.Command](https://www.openhab.org/javadoc/latest/org/openhab/core/types/command) | `[item] received a command`            | Command that triggered event                                                                                  | `receivedCommand`      |     |
+| `itemName`     | string                                                                                                               | all                                    | Name of Item that triggered event                                                                             | `triggeringItem.name`  |     |
+| `type`         | string                                                                                                               | all                                    | Type of event that triggered event (`"ItemStateEvent"`, `"ItemStateChangedEvent"`, `"ItemCommandEvent"`, ...) | N/A                    |     |
+
+Note that in UI based rules `event.itemState`, `event.oldItemState`, and `event.itemCommand` are java types (not javascript), and care must be taken when comparing these with javascript types:
```
florian-h05, line 154: Spell `java` uppercase (`Java`) and spell `javascript` as ``JavaScript` (two times).

florian-h05 (changes requested):
Overall looks really well, just a few spelling and formatting points.

florian-h05:
> is @runtime available in file based rules? If not, it should be mentioned in docs

Yes, it is also available in file-based scripts/rules.

README.md
```diff
@@ -767,6 +904,9 @@ rules.when().item('HallLight').receivedUpdate().then().copyState().fromItem('Bed
 ```
 
 ### Event Object
+
+**NOTE**: This section is only valid for File Based Rules. If you are using MainUI with jsscripting
```
florian-h05, line 908: This seems incomplete, I think you forgot something here.

I would write:

> **NOTE**: This section is only valid for file-based scripts/rules. If you use JS Scripting in MainUI, please refer to ...

README.md
```diff
@@ -15,41 +15,68 @@ It provides convenient access to common openHAB functionality within rules inclu
...
-  - [Actions](#actions)
-  - [Cache](#cache)
-  - [Log](#log)
-  - [Time](#time)
-  - [Utils](#utils)
-- [File Based Rules](#file-based-rules)
-  - [JSRule](#jsrule)
-  - [Rule Builder](#rule-builder)
-  - [Event Object](#event-object)
+- [openHAB Javascript Library](#openhab-javascript-library)
```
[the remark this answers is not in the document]
ssalonen: https://github.com/openhab/openhab-js/pull/110/commits/023d2849aa996bdcbf016ab679786e6edc832d9a

README.md
ssalonen: https://github.com/openhab/openhab-js/pull/110/commits/7cfd0822b42145bd1cd201aa646042087cc3914b

README.md
ssalonen: https://github.com/openhab/openhab-js/pull/110/commits/fb6bbf5f7f4b7ddcb308f4a1798ad3ff47f8853a

README.md
ssalonen: https://github.com/openhab/openhab-js/pull/110/commits/a49c8d9f526685048d7f64f6a1017ae48ef1e40c

README.md
ssalonen: https://github.com/openhab/openhab-js/pull/110/commits/9315b918b593c1eca955ec1f6ddbc37e6843733e

README.md
ssalonen: https://github.com/openhab/openhab-js/pull/110/commits/9315b918b593c1eca955ec1f6ddbc37e6843733e

florian-h05 (approved):
LGTM👍  

Thank you @ssalonen.

florian-h05:
We need @digitaldan or @jpg0 for merging.

florian-h05:
@ssalonen 
It might be possible that your event docs are removed soon as  I am currently working on unifying the event object in UI and file-based rules.

README.md
```diff
@@ -41,15 +43,19 @@ binding](https://www.openhab.org/addons/automation/jsscripting/)
...
 
-- Install the openHAB [JavaScript binding](https://www.openhab.org/addons/automation/jsscripting/), a version of this
-library will be automatically installed and available to all javascript rules by default.
+By default, openHAB has fairly old JavaScript runtime (in MainUI `ECMA - 262 Edition 5.1` or `application/javascript`). In [community forum](https://community.openhab.org) many people also call this as "Nashorn engine".
+
+This library is not compatible with this old runtime.
 
+Instead, this library is meant to be used with ECMAScript 2021+ engine (also known as `ECMAScript 262 Edition 11`, `application/javascript;version=ECMAScript-2021` or `jsscripting` addon). Rules can be created using files ([File Based Rules](#file-based-rules)) or MainUI ([UI Based Rules](#ui-based-rules)). 
+
+By installing the openHAB [JavaScript binding](https://www.openhab.org/addons/automation/jsscripting/), a version of this library will be automatically installed and available to ECMAScript 2021+ rules by default.
```
digitaldan, line 58:
```suggestion
Install the openHAB [JavaScript binding](https://www.openhab.org/addons/automation/jsscripting/), a version of this library will be automatically installed and available to ECMAScript 2021+ rules created using [File Based Rules](#file-based-rules) or [UI Based Rules](#ui-based-rules). 

By default, openHAB ships with an older JavaScript runtime based on the Nashorn JavaScript engine which is part of the standard JDK.  This is referred to as `ECMA - 262 Edition 5.1` or `application/javascript` in the Main UI.

*This library is not compatible with this older runtime.* 
```

digitaldan (changes requested):
Thanks!  Looks good, i had only one suggestion

README.md
```diff
@@ -117,6 +155,82 @@ The openHAB JSScripting runtime attempts to provide a familiar environment to Ja
 Scripts may include standard NPM based libraries by using CommonJS require.
 The library search path will look in `automation/js/node_modules` in the user configuration directory.
 
+### @runtime
```
digitaldan, line 158: So , i think this is super helpful, but it's also a lot to put in the "Scripting Basics" section which was meant to be more of a quick overview of commonly used features.  Is there somewhere else we can put this farther down?  My fear is that new users to this library will get a little overwhelmed with the large table of runtime objects and miss the other basics .

florian-h05: What about introducing a new section named like `Advanced Scripting` or `For Advanced Users` at the bottom of the README?

florian-h05:
@digitaldan 
What about using _Squash and Merge_ instead of just merging in this and any other future PR?
addons and core seem to squash and merge and this keeps commit history very clean.

digitaldan:
Yeah, i have been thinking this was happening already, but forgot repos defaults to not squashing, i just changed the settings so now squashing always happens, thanks for noticing !

README.md
ssalonen: I would respectfully disagree here.

As you know, the stuff from @runtime is available for basic use with the rules dsl and nashorn, and I would argue it is as commonly used/needed with ecmascript.

In my opinion you often need to access globals from there, and it is really hard to stay in pure js world without some sort of interaction with Java classes (QuantityType, ON, OFF, ZonedDateTime), would you not agree?

In fact, with my limited experience, you really need to pay attention to types since the comparisons work perhaps unlike one would expect (I would expect "dummy" == StringType("dummy") to equal True for example. Is this graalvm issue or just general javascript?)

florian-h05: You are totally right that that part mentions important points, but personally I would still consider it as advanced as the normal user should use openhab-js and not directly access the Java parts.

I think I am more than just a normal user and I have plenty of rules in JavaScript and I do not use @runtime anywhere as I do not need to, instead I exclusively use openhab-js to interact with the openHAB runtime.

ssalonen: https://github.com/openhab/openhab-js/pull/110/commits/667da0817b8d584c30347d30451b4dafeadd1fc9

ssalonen:
I took the liberty of moving init/deinit hook sections: https://github.com/openhab/openhab-js/pull/110/commits/9f038989915ac4e8f426c29b3c41afa62620e6d1

digitaldan:
@ssalonen this looks great, thanks!  I like the advanced scripting section, i can see additional parts being added to that over time.

digitaldan:
Can you fix the conflicts that have popped up due to other PR merges?

ssalonen:
@digitaldan should be now OK. PR #99  caused a lot of issue for rebase, it seemed to break havoc to many parts of the doc... Best to verify diff of this PR to ensure it is OK

ssalonen:
> > is @runtime available in file based rules? If not, it should be mentioned in docs
> 
> I didn't see an answer to this. You can `require(@runtime)`, however there are conflicts. For example, @runtime defines `items` to be a java.util.Map of Item names and their current state whereas openhab-js defines `items` to be your interface for doing all things related to Items and the Item registry and such. 

It was confirmed that it is also in file-based rules.

> So if you are using openhab-js, you probably shouldn't be using `@runtime`.

Personally I do not agree with this. Without `@runtime` it is impossible to construct `QuantityTypes` and compare values, right?

[Anecdotally](https://github.com/openhab/openhab-js/pull/110#discussion_r863061474), you can still do plenty without `@runtime`.

florian-h05:
Of course the use of @runtime can cause conflicts, but it is listed under advanced scripting and there is a note that the usage of openhab-js instead is recommended.
For some use cases, e.g. `QuantityType`, openhab-js is type of incomplete now and I think it is great to document @runtime even though openhab-js is recommended over it.

> Without @runtime it is impossible to construct QuantityTypes and compare values, right? Or sure, you can import your QuantityType using Java.type as well.

That would be one-thing that we should improve in openhab-js.

rkoshak:
> Without @runtime it is impossible to construct QuantityTypes and compare values, right?

No, you don't need the entirety of @runtime to work with QuantityTypes. And I've already opened an issue and started experimenting with ways to manage QuantityTypes in the library where possible (see #102) . Lack of operator overloading will make it challenging though. But ideally I'd like users to be able to compare and do math with numbers and QuantityTypes interchangeably without having to worry about units unless they want to.

Anyway, you have two choices. Option one is to throw out the units:

```
if(items.getItem('MyTemp').rawState.floatValue() > 65)
```

You can event force it to °F if you need to.

```
if(items.getItem('MyTemp').rawState.toUnit('°F').floatValue() > 65)
```

No imports are required at all with this approach but, of course, it works best when you are working with constants and such. If you are working with to QuantityTypes it's probably best not to throw out the units. It also requires you to know ahead of time that the Item is a Number:X. 

But yes, you'd have to either `Java.type` it or just import QuantityType from @runtime, not all of @runtime to work strictly in QuantityTypes.

```
var QT = Java.type('org.openhab.core.library.types.QuantityType'); // or

if(items.getItem('MyTemp').rawState.compareTo(new QT(65, '°F')) > 1)
```

Importing just QuantityType from @runtime would look something like the following I think.

```
var {QuantityType} = require('@runtime');
```

I've no problem with documenting `@runtime` in the library at all. But I, like @florian-h05, do not use `@runtime` in any of my rules. In those exceptionally rare cases where I need something that is normally in `@runtime`, I `Java.type` them or import just those from `@runtime`. It will provide a much smoother experience for end users to not use `@runtime` unless they really know what they are doing because of the naming conflicts. The library also goes out of its way to wrap and abstract away from the raw openHAB Java Objects to provide a pure(ish) JavaScript environment whereas `@runtime` gives you raw Java which causes hard to understand (for many users) type problems. 

So I agree it's an advanced topic and not something most users should be using as a kind of default import.

jpg0:
A rule of thumb as to when `@runtime` is required is when you need an instance of an osgi service from inside openHAB. You do not need it to import classes (they are not instances, nor inside the osgi container). You would need it to get hold of the various providers, registries etc.

rkoshak:
> A rule of thumb as to when @runtime is required is when you need an instance of an osgi service from inside openHAB

Wouldn't you want to use `osgi.lookupService()` or `osgi.getService()` for that? That's what I've been using.

jpg0:
> Wouldn't you want to use `osgi.lookupService()` or `osgi.getService()` for that? That's what I've been using.

Sorry, yes, you're right. I forgot that I added this to the JS layer. I guess the only remaining reason is for anything which the devs have (a) added to `@runtime` but (b) not exposed in the JS layer. I hope this trends to nothing, leaving no remaining reason to use `@runtime`.

ssalonen:
`@runtime` offers important migration path from Nashorn, as guided by @rkoshak  https://community.openhab.org/t/migrate-from-nashorn-to-jsscripting-for-ui-rules/129788

> ```
> if(items.getItem('MyTemp').rawState.compareTo(new QT(65, '°F')) > 1)
> ```

Personally, (and this is just me personally!), I prefer to have the framework protect against incompatible units and such. 

Don't get me wrong, I am all for 'native' js adaptation of core java types, and have those documented from a common place. I can imagine the current setup is jarring for new users as you need to lookup and combine documentation from many places.

PR #116, opened 2022-05-17
Rules: Failed rule run logs a useful error message

### LINKED ISSUE #115: Rules: Logged error on failed rule run doesn‘t contain any error information
## Expected Behavior
I would expect that the logged message includes a detailed error message like the following:
```
org.openhab.core.items.ItemNotFoundException: Item 'Aussenlicht' could not be found in the item registry
```

## Current Behavior
When a JSRule run failes, a message like that is logged:
```
2022-05-16 22:31:24.256 [ERROR] [e.automation.internal.RuleEngineImpl] - Failed to execute rule 'Test-Rule-2c1e38ee-dec4-432f-b0ab-cd3192361c04': Fail to execute action: 1
```

## Possible Solution

I think I already have some solution, but this required additional testing.
I will soon open a PR.

1. Create a JSRule that will produce an error, e.g. by using `items.getItem()` on a non existing Item.
2. Run the rule.

## Your Environment

* Version used: (e.g., openHAB and **JS Scripting add-on version**): openHAB 3.3.0.M5
* openhab-js version used: 1.2.3

florian-h05:
Fixes #115.

When a rule run fails, now the JavaScript error is logged (instead of the useless `Failed to execute action: 1` from before), which looks like:
```
2022-05-17 13:59:21.285 [ERROR] [enhab.automation.script.file.test.js] - Failed to execute rule Test-Rule-73bd4126-38b7-4ce7-97bb-98aa4cafa0b5: org.openhab.core.items.ItemNotFoundException: Item 'Aussenlicht' could not be found in the item registry
```

### DIFF
--- a/rules/rules.js
+++ b/rules/rules.js
@@ -212,13 +212,14 @@ let JSRule = function (ruleConfig) {
       } catch (error) {
         // logging error is required for meaningful error log message
         // when throwing error: error is caught by core framework and no meaningful message is logged
+        let msg;
         if (error.stack) {
-          console.error(`Failed to execute rule ${ruid}: ${error}: ${error.stack}`);
-          throw Error(`Failed to execute rule ${ruid}: ${error}: ${error.stack}`);
+          msg = `Failed to execute rule ${ruid}: ${error}: ${error.stack}`;
         } else {
-          console.error(`Failed to execute rule ${ruid}: ${error}`);
-          throw Error(`Failed to execute rule ${ruid}: ${error}`);
+          msg = `Failed to execute rule ${ruid}: ${error}`;
         }
+        console.error(msg);
+        throw Error(msg);
       }
     };
 
--- a/rules/rules.js
+++ b/rules/rules.js
@@ -214,8 +214,10 @@ let JSRule = function (ruleConfig) {
         // when throwing error: error is caught by core framework and no meaningful message is logged
         if (error.stack) {
           console.error(`Failed to execute rule ${ruid}: ${error}: ${error.stack}`);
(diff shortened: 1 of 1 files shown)


jpg0:
Great work, this has annoyed me for a long time!

florian-h05:
Yeah, this makes script debugging possible from now on.
Can you merge?

jpg0:
The only question I have regarding merging is whether swallowing the exception (rather than rethrowing it after logging it) causes any other changes. For example does this mean that openHAB would think that the rule ran successfully even though it didn't? In that case would it be better to log and rethrow?

florian-h05:
You are absolutely right with your thoughts, I do not know whether this will cause any problems, so I will rethrow.

rules/rules.js
```diff
@@ -207,11 +207,19 @@ let JSRule = function (ruleConfig) {
...
+        return ruleConfig.execute(getTriggeredData(input));
+      } catch (error) {
+        // logging error is required for meaningful error log message
+        // when throwing error: error is caught by core framework and no meaningful message is logged
+        if (error.stack) {
+          console.error(`Failed to execute rule ${ruid}: ${error}: ${error.stack}`);
+          throw Error(`Failed to execute rule ${ruid}: ${error}: ${error.stack}`);
+        } else {
+          console.error(`Failed to execute rule ${ruid}: ${error}`);
+          throw Error(`Failed to execute rule ${ruid}: ${error}`);
```
jpg0, line 220: Would it be cleaner to:

```
let msg = `Failed to execute rule ${ruid}: ${error} ${error.stack ? ": " + error.stack : "" }`
console.error(msg);
throw Error(msg);
```

So that you don't need to repeat the message so many times?

PR #125, opened 2022-05-28
[rules] Improvements for event object & JSRule callback

### LINKED ISSUE #117: PWM / PID controller rule from template in js syntax
Hi, i need your help with syntax for PWM and / or PID controller rule creation using JSRule. 

1. My goal is to create such rule (snippset of code after creation in GUI) using JSRule
```
configuration: {}
triggers:
  - id: aa3cd1af-21b2-4ea2-b04b-bdcb4912ad93
    label: PWM Trigger
    configuration:
      deadManSwitch: 50000
      dutycycleItem: HeaterSwitch_dutycycle
      minDutycycle: 0
      interval: 600
      maxDutycycle: 100
    type: pwm.trigger
conditions: []
actions:
  - inputs: {}
    id: "1"
    configuration:
      itemName: HeaterSwitch
    type: core.ItemCommandAction
  - inputs: {}
    id: "2"
    configuration:
      type: application/javascript
    type: script.ScriptAction

```
2. My first attemp is: 
rules.JSRule({
  name: "Termostat x PWM rule",
  id:   "Termostat_x_PWM",
  tags: ['Roomx','_DYNAMIC_'],
  overwrite: "true",
  description: "PWM rule for heating actuator",
  triggers: [triggers.pwm('RadiatorACx_homekit_TargetHeatingCoolingMode')],
  execute: data => {
    
  }
of course it is now working as expected. 
Can somebody help me how to define the triggers with items in configuration

### LINKED ISSUE #124: [Docs] File Based Rules/Event Object types are incorrect
Recently I found out that the type of the `receivedCommand` property is not string as I wrote in the docs.

I will take a look into the types of the event object‘s properties and correct the docs.

florian-h05:
This PR fixes #124.

## Enhancements
-  `JSRule execute` callback: Allow access to raw Java input.
- `event` object: Make all properties typeof `string`.
- Document both callback and `event` object in JSDoc.

## Possible breaking changes
- The `receivedCommand` property of the `event` object is now typeof `string`, before it was the raw Java command type.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,7 +13,7 @@ This library aims to be a fairly high-level ES6 library to support automation in
 It provides convenient access to common openHAB functionality within rules including items, things, actions, logging and more.
 
 This library is included by default in the openHAB [JavaScript
-binding](https://www.openhab.org/addons/automation/jsscripting/)
+binding](https://www.openhab.org/addons/automation/jsscripting/).
 
 - [Installation](#installation)
   - [Default Installation](#default-installation)
@@ -741,7 +741,7 @@ rules.when().item("F1_light").changed().then(event => {
     console.log(event);
 }).build("Test Rule", "My Test Rule");
 ```
-see [Examples](#rule-builder-examples) for further patterns
+See [Examples](#rule-builder-examples) for further patterns.
 
 #### Rule Builder Triggers
 
@@ -862,13 +862,13 @@ This tables gives an overview over the `event` object:
 | `eventType`       | all except `ThingStatus****Trigger`s                | Type of event that triggered event (change, command, time, triggered, update)       | N/A                    |
 | `triggerType`     | all except `ThingStatus****Trigger`s                | Type of trigger that triggered event (for `TimeOfDayTrigger`: `GenericCronTrigger`) | N/A                    |
 
-All properties are type of String.
+All properties are typeof `string`.
 
 **NOTE:**
 `ThingStatusUpdateTrigger`, `ThingStatusChangeTrigger` use *Thing* and `ChannelEventTrigger` uses the the trigger channel name as value for `itemName`.
 `Group****Trigger`s use the equivalent `Item****Trigger` as trigger for each member.
 
-You may use [utils.dumpObject(event)](https://openhab.github.io/openhab-js/utils.html#.dumpObject) to get all properties of an `event` object.
+See [openhab-js - EventObject](https://openhab.github.io/openhab-js/rules.html#.EventObject) for full APi documentation.
 
 ### Initialization hook: scriptLoaded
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -849,26 +849,7 @@ When a rule is triggered, the script is provided the event instance that trigger
...
-| `triggerType`     | all except `ThingStatus****Trigger`s                | Type of trigger that triggered event (for `TimeOfDayTrigger`: `GenericCronTrigger`) | N/A                    |
-
-All properties are type of String.
-
-**NOTE:**
-`ThingStatusUpdateTrigger`, `ThingStatusChangeTrigger` use *Thing* and `ChannelEventTrigger` uses the the trigger channel name as value for `itemName`.
-`Group****Trigger`s use the equivalent `Item****Trigger` as trigger for each member.
-
-You may use [utils.dumpObject(event)](https://openhab.github.io/openhab-js/utils.html#.dumpObject) to get all properties of an `event` object.
+See [openhab-js - EventObject](https://openhab.github.io/openhab-js/rules.html#.EventObject) for documentation.
```
digitaldan, line 852: So we have the table describing the event object when using UI based rules earlier in this file, and then this one which describes the event object when using files.  Is there a reason to remove this one, but not the other?  I think its nice to have here as knowing what's in the event object is pretty important and personally i ended up doing a console.dir(event) for a while in my rules to remember what was being passed in.

florian-h05: I have removed the file-based event object table from the README and instead linked the new file-based event object doc in JSDoc.

florian-h05: So the docs are not gone but moved.

digitaldan: Right , I see that, but now we list the event object only for UI rules but not files? It seems like we should have both or none in the README.   My other point is that i see this as information anyone who is using this library will need, so maybe its best to have it in the readme vs having to read the API docs which is i think is for more advanced users?

florian-h05: I see your point, I will update the file based one in the README instead of removing it.
I will keep the link to JSDoc dir more complete docs, but add back the basics to the README.

rules/rules.js
```diff
@@ -209,7 +237,7 @@ const JSRule = function (ruleConfig) {
 
   const doExecute = function (module, input) {
     try {
-      return ruleConfig.execute(getTriggeredData(input));
+      return ruleConfig.execute(getTriggeredData(input), input, module);
```
digitaldan, line 240: I see #117 needs the command from the input, should we modify getTriggeredData to return an object that has the needed command as a string member?  I'm always a bit hesitant to overload calls like this if we don't have to ( the api starts to bloat)

README.md
digitaldan: 👍

digitaldan:
LGTM, thanks!

PR #126, opened 2022-05-28
[triggers] Add PWM trigger & Enhance JSDoc for triggers

### LINKED ISSUE #117: PWM / PID controller rule from template in js syntax

florian-h05:
Depends on #125.
Fixes #117.

This PR introduces a triggers for the PWM automation & an example PWM rule (the example depends on #125).

It furthermore enhances the JSDoc for the triggers (I had to make all triggers const and then export them instead of directly exporting them!!).

### DIFF
--- a/triggers.js
+++ b/triggers.js
@@ -32,263 +32,262 @@ const createTrigger = function (typeString, name, config) {
     .build();
 };
 
-module.exports = {
+/**
+ * Creates a trigger that fires upon specific events in a channel.
+ *
+ * @example
+ * ChannelEventTrigger('astro:sun:local:rise#event', 'START')
+ *
+ * @memberof triggers
+ * @param {String} channel the name of the channel
+ * @param {String} event the name of the event to listen for
+ * @param {String} [triggerName] the optional name of the trigger to create
+ *
+ */
+const ChannelEventTrigger = (channel, event, triggerName) => createTrigger('core.ChannelEventTrigger', triggerName, {
+  channelUID: channel,
+  event: event
+});
 
-  /**
-     * Creates a trigger that fires upon specific events in a channel.
-     *
-     * @example
-     * ChannelEventTrigger('astro:sun:local:rise#event', 'START')
-     *
-     * @name ChannelEventTrigger
-     * @memberof triggers
-     * @param {String} channel the name of the channel
-     * @param {String} event the name of the event to listen for
-     * @param {String} [triggerName] the optional name of the trigger to create
-     *
-     */
-  ChannelEventTrigger: (channel, event, triggerName) => createTrigger('core.ChannelEventTrigger', triggerName, {
-    channelUID: channel,
-    event: event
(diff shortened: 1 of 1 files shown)


digitaldan:
Huh, i was not actually aware of custom triggers like this , looking through the addons code base, it seems like PWM, PIDController and Hue Emulation or the only addons that support implement custom triggers ATM.  I was wondering if there was a way to dynamically add triggers found  in the OSGI registry in openhab-js so we don't have to keep up with new bindings, but given the small number of bindings that use it, I guess i'm not that concerned about it.

Code looks good, thanks!

florian-h05:
@digitaldan 
I also wasn’t aware of those custom tiggers until issue #117 was opened.
I have holidays soon and then I‘ll probably add the triggers for PIDController and Hue Emulation.

digitaldan:
Sounds good, the PID controller looks useful, the Hue Emulation triggers are not  actually documented, and i'm not sure are named correctly, instead of being something like `hueemulation.TimerTrigger` its `timer.TimerTrigger`, which does not sound right to me, see:

https://github.com/openhab/openhab-addons/blob/4ded5732c33ea25d64039d728bf39a4615885e00/bundles/org.openhab.io.hueemulation/src/main/resources/OH-INF/automation/moduletypes/TimerTrigger.json#L4

We should probably open a ticket up clarifying if that's intentional if we decide to add it, but I'm not sure if anyone actually does  use it, so i would say its a low priority.

florian-h05:
You are right, PID sounds really useful.

Regarding Hue emulation, as the triggers are not documented I would not add them until somebody explicitly asks to, so yes, low priority.

PR #134, opened 2022-06-06
[rules] Fix EventObject.itemName `null` is `string` instead of `null`

florian-h05:
Fixes an issue that I introduced in #125.

I added `toString()` to event.itemName, but this made null checks fail because even null is now a string. I double checked, and itemName is also string if I remove `toString()`.


florian-h05:
@digitaldan Can you review?

digitaldan:
👍

florian-h05:
@digitaldan 
I know that I am too late, but I found out that the problem was not only my previous PR #125, but the problem is in the `getTriggeredData` function itself, therefore I am currently doing further investigation and I will refactor that method completely.

PR #143, opened 2022-06-27
[triggers] Update documentation

### LINKED ISSUE #140: [triggers] How to skip parameter in trigger configuration
https://openhab.github.io/openhab-js/triggers.html#.ItemStateChangeTrigger says:

> (static) ItemStateChangeTrigger(itemName, oldStateopt, newStateopt, triggerNameopt)

That is, the `oldState`, `newState` and `triggerName` parameters are optional.  Please document how to skip `oldState` and utilize only `newState`.

I want to crate a trigger, only based on the new state, irrespective of the old state.  The concrete case is, that I have broken inverter and sometimes it resets and produces `0 W`atts.  I want to trigger an event, when the watts are zero, irrespective of the previous state of the watts.

florian-h05:
Fixes #140.

Update JSDoc to include more examples and to skip parameters.


digitaldan:
LGTM, thanks!

PR #148, opened 2022-07-10
[actions] Extend type definitions

florian-h05:
This extends (or really enables) the type definitions for the `actions` namespace.

I needed to change the way of exporting the actions to have the type definitions generated.

Also, I added the NotificationAction using a try_catch statement to cover the (rare?) case that somebody does not have the Cloud Connector add-on installed. 

@digitaldan This is the PR I mentioned yesterday.


florian-h05:
@digitaldan 
I am planning to publish a new version (2.0.1) soon after this PR gets merged to publish the two bugfixes I made as well as this enhancement.

actions.js
```diff
@@ -253,10 +261,51 @@ const ThingsAction = Java.type('org.openhab.core.model.script.actions.Things');
...
+ * NotificationAction.sendLogNotification('<message>'); // only listed in the notification log
+ *
+ * @name NotificationAction
+ * @memberof actions
+ */
+let NotificationAction;
+try {
+  NotificationAction = Java.type('org.openhab.io.openhabcloud.NotificationAction');
+} catch (error) {
+  if (error.name !== 'TypeError') throw Error(error);
```
digitaldan, line 284: I'm assuming if the cloud binding is not installed it will throw this 'TypeError' ?

digitaldan (approved):
LGTM!

actions.js
florian-h05: Yes.
When the binding is not installed, the loading of the class fails with the TypeError.

florian-h05: This error is caught, other errors are rethrown.

PR #151, opened 2022-08-16
[time] Fix `isBetweenTimes` fails with `time.toZDT` not a function

### LINKED ISSUE #150: time.isBetweenTimes cannot see time.toZDT()
## Expected Behavior

When calling `myZDT.isBetweenTimes(before, after)` I expect to get back a boolean.

## Current Behavior

An error is thrown:

    an error occured: TypeError: time.toZDT is not a function

I know this worked when I checked it in because I had test cases to exercise it. But I can't see anything that has changed in time.js since then to cause it to break. 

## Possible Solution

I don't understand the problem yet so I've no idea how to fix it. Clearly there is something going on that is making it invisible to the isBetweenTimes function.

1. Create a ZonedDateTime. I use `time.toZDT()` just to prove to myself that there is nothing wrong with `time.toZDT()`. Let's name it `myTime`.
2. Call `myTime.isBetweenTimes("06:00:00", "12:00:00")` or just about anything that `toZDT` supports.
3. See the error in the logs.

## Your Environment

* Version used: (e.g., openHAB and **JS Scripting add-on version**): 3.4.0 #3049
* Environment name and version (e.g. Chrome 76, Java 8, Node.js 12.9, ...): Official Docker Image
* Operating System and version (desktop or mobile, Windows 10, Raspbian Buster, ...): Official Docker Image (Debian variant)

florian-h05:
@rkoshak This is the PR I was talking about in the issue.

Fixes #150.
Previous PR #137 changed the way how toZDT is defined, which broke
isBetweenTimes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 | Type        | Namespace | Description                                                              | Reference                                      | Breaking |
 |-------------|-----------|--------------------------------------------------------------------------|------------------------------------------------|----------|
+| Bugfix      | `time`    | `isBetweenTimes` fails with: TypeError: `time.toZDT` is not a function   | https://github.com/openhab/openhab-js/pull/151 | No       |
 
 Also see the [Release Milestone](https://github.com/openhab/openhab-js/milestone/3).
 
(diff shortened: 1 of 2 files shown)


rkoshak (commented):
I tested it and it works for me. Thanks for the quick fix!

PR #161, opened 2022-09-28
Update timer docs

florian-h05:
This improves the docs for `actions.ScriptExecution` Timers by explaining how to pass variables to the callback function.

### DIFF
--- a/README.md
+++ b/README.md
@@ -526,7 +526,7 @@ Replace `<url>` with the request url.
 See [openhab-js : actions.ScriptExecution](https://openhab.github.io/openhab-js/actions.html#.ScriptExecution) for complete documentation.
 
 ```javascript
-let now = time.ZonedDateTime.now();
+var now = time.ZonedDateTime.now();
 
 // Function to run when the timer goes off.
 function timerOver () {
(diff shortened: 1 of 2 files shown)


florian-h05:
@rkoshak 
Would be nice if you could have a quick look at the docs changes.

rkoshak:
It looks good to me. The only thing I would add is the generic case of just passing one local variable to the timer instead of `this`. Theoretically (I've never done it though) that too would fix the variable same as the function generator does. But you can only pass the one variable.

florian-h05:
> The only thing I would add is the generic case of just passing one local variable to the timer instead of this. 

That‘s possible, but …

> that too would fix the variable same as the function generator does. But you can only pass the one variable.

I expect that it is not the same as with the function generator, I think you can mutate the var after the creation of the timer. 
I have to test this to check my assumption. 

I will modify the docs to have two ways of passing a var to a timer, both ways support `this`. The main difference should be that in one case the var can be changed after timer creation.

florian-h05:
@rkoshak 
Ready for re-review.

rkoshak:
Once that comment is added I think this is good to go.

PR #167, opened 2022-10-20
[items] Add semantics to the `Item` class

### LINKED ISSUE #162: [items] Add a "rawItem" attribute to Item similar to rawState
There are rare but useful circumstances where getting access to the raw Java Item Object is required. One example is when using the Semantic Actions (e.g. getEquipment(), getLocation(), etc.). These Actions require the Java Item Object and, at least based on my experience, there is nothing to autotranslate between the JS Item and Java Item like there is for ZonedDateTime.

So my thoughts are:

1. Simply add a `rawItem` attribute/setter/getter to Item that returns the original Java Item Object. This will give the user access to it when it's needed without needing to import the ItemRegistry and pulling it from there.
2. Add the magic to the add-on that auto translates between the Java and JS like is done for ZonedDateTime (if that's even possible)
3. Add the semantic actions to the Item class itself similar to ItemHistory which solves my immediate problem but leaves open problems where the Java Item is required elsewhere that I don't know about yet.

I can implement 1 or 3 myself but I think 2 needs to be moved to the add-on. 

Thoughts? What feels like the best approach. I *really* don't like needing to import the ItemRegistry to access these calls.

florian-h05:
@rkoshak 
Fixes #162.

## Improvements to the `Item` class

- Documented the `rawItem` attribute
- Added a `semantics` attribute that is holding the new `ItemSemantics` class
- Added the `ItemSemantics` class that provides access to some semantic information.
- Fixed the intendation of the JSDoc comments.

## Testing

I have tested those changes by editing an Item and then running the following script:

```javascript
var sem = items.getItem('test_switch').semantics;

console.info(sem.equipmentType);
console.info(sem.locationType);
console.info(sem.pointType);
console.info(sem.propertyType);
console.info(sem.semanticType);
```


rkoshak:
Looks good. This is going to simplify a bunch of my rules which use the semantic model.

florian-h05:
@rkoshak 
Can I merge from your side?

I think that @digitaldan is busy, and if you give approval I think that I can merge this.

rkoshak:
Yes, it looks good to merge from my side. 

But I'm not a reviewer on this repo so I can't do anything to officially mark off the review. If we need more than just this comment I'm afraid we'll have to wait for Dan.

florian-h05:
> But I'm not a reviewer on this repo so I can't do anything to officially mark off the review. If we need more than just this comment I'm afraid we'll have to wait for Dan.

Somewhere it is written that a maintainer should not merge his/her own PRs, but instead one of the other maintainers should merge.
Practically, when looking at some of the openHAB repos, some maintainers merge their own PRs.

This one has your review (yes, we can't mark it of), so in my opinion we could merge, but I also have nothing against waiting for Dan.

digitaldan (approved):
Hi, i don't use semantics much, but the actual code LGTM.  I did not run through the readme changes carefully just FYI.

PR #169, opened 2022-11-05
Update to addon changes

florian-h05:
Fixes #168.

## Changes

- Update docs & workflows for NodeJS v16.17.1
- Update package-lock for NodeJS v16.17.1
- README: Mention that the timer polyfills behave now like standard JS
- README: Remove the openHAB Timer and the raw timer creation methods
- README: Add a warning for changing logger names
- README: Explain access to vars in the standard JS timer methods (e.g. function generator)
- `actions`: Log warning when the raw timer creation methods are accessed
- Clean & Upgrade some dev dependencies

## Testing

I have also verified that the library is still working under older versions of the addon that are running an older GraalJS and therefore NodeJS version.

Use the following script to verify all relevant changes:

```javascript
const { actions, time } = require('openhab');

// Test whether access to raw timer creation methods is still possible but logs a warning
const now = time.ZonedDateTime.now();

actions.ScriptExecution.createTimer('timer1', now.plusSeconds(2), () => {
  console.info('Timer with identifier ran');
});
actions.ScriptExecution.createTimer(now.plusSeconds(4), () => {
  console.info('Timer without identifier ran');
});

const foo = 'myVar';

actions.ScriptExecution.createTimerWithArgument('timer2', now.plusSeconds(6), foo, () => {
  console.info('Timer with identifier and var ran: ' + foo);
});

actions.ScriptExecution.createTimerWithArgument(now.plusSeconds(8), foo, () => {
  console.info('Timer without identifier and with var ran: ' + foo);
});

// Test examples from the README updates

// Example 1

var myVar = 'Hello world!';

// Schedule a timer that expires in ten seconds
setTimeout(() => {
  console.info(`Timer expired with myVar = "${myVar}"`);
}, 10000);

myVar = 'Hello mutation!'; // When the timer runs, it will log "Hello mutation!" instead of "Hello world!"

// Example 2

// Function generator for the timer's callback function
function cbFuncGen (myVariable) {
  return function () {
    console.info(`Timer expired with myVar = "${myVariable}"`);
  };
}

var bar = 'Hello world!';

// Schedule a timer that expires in ten seconds
setTimeout(cbFuncGen(bar), 12000);

bar = 'Hello mutation!'; // When the timer runs, it will log "Hello world!"
```

PR #171, opened 2022-11-12
[actions] Don't deprecate `ScriptExecution.createTimer`

florian-h05:
Reference https://community.openhab.org/t/js-scripting-why-are-we-deprecated-createtimer/140748.
Refercence https://github.com/openhab/openhab-addons/pull/13695.

## Description

This PR reverts the deprecation of `createTimer` from previous PR https://github.com/openhab/openhab-js/pull/169.

As discussed on the forum https://community.openhab.org/t/js-scripting-why-are-we-deprecated-createtimer/140748, the setTimeout and setInterval methods do not provide the functionality that advanced users need, and to avoid that those users and users coming from Rules DSL leave JS Scripting because of a missing advanced timer creation API, `createTimer` is reimplemented to be thread-safe in https://github.com/openhab/openhab-addons/pull/13695.

`createTimerWithArgument` is still deprecated, as there is no use case for it in JS (see https://github.com/openhab/openhab-addons/pull/13695#discussion_r1020755441).

This PR is backward compatible and does not require the addon to be on the latest add-on version with the thread-safe reimplementation of `createTimer`, but keep in mind that if not, this can lead to multithread errors.

## Testing

Use the following file-based script:

```javascript
const { actions, time } = require('openhab');

const now = time.toZDT();

// Note that all timers are scheduled to nearly the same time, so without the synchronization mechanism in place, this would lead to a multi-thread exception
actions.ScriptExecution.createTimer(now, () => {
    console.info("Hello world from createTimer without identifier")
});

actions.ScriptExecution.createTimer('createTimer', now, () => {
    console.info("Hello world from createTimer with identifier")
});

let arg1 = 'test';
actions.ScriptExecution.createTimerWithArgument(now, arg1, (newArg) => {
    console.info(`Hello world from createTimerWithArgument without identifier and argument: ${newArg}`);
});

actions.ScriptExecution.createTimerWithArgument('createTimerWithArgument', now, arg1, (newArg) => {
    console.info(`Hello world from createTimerWithArgument with identifier and argument: ${newArg}`);
});
```


florian-h05:
@rkoshak Could you possibly review at least the README changes?

digitaldan (approved):
LGTM

PR #172, opened 2022-11-15
Downgrade NodeJS version & Improve markdownlint workflow

florian-h05:
Reference https://github.com/openhab/openhab-addons/pull/13714


florian-h05:
@digitaldan After I had to downgrade GraalJS in JS Scripting, the NodeJS version changed again, now to 14.18.1 — I also downgraded our library.
But I am wondering, whether it is required or should be the case that the library is developed on the same NodeJS version that the addon is using, or if it is completely okay to develop the library on NodeJS 16 and use it on 14.

florian-h05:
Replaced by https://github.com/openhab/openhab-js/pull/178.

PR #173, opened 2022-11-18
[WIP] [rules] Add support for displaying the code in  MainUI

florian-h05:
Fixes https://github.com/openhab/openhab-addons/issues/12929.

## Description

With this PR, the `JSRule` API adds additional information to the rule‘s action, so that MainUI is able to display a rule‘s source code.
Due to the lack of direct access to the rules source code, a utility function is used to regenerate the source code from the given rule configuration.
Therefore, the generated code does not work for the `RuleBuilder` API, as this is using `JSRule` under the hood.

## Testing

This PR requires additional testing from my side, as well as an update of the type defintions.

### DIFF
--- a/rules/rules.js
+++ b/rules/rules.js
@@ -283,6 +283,11 @@ const JSRule = function (ruleConfig) {
     rule = registerRule(rule);
   }
 
+  // Add config to the action so that MainUI can show the script
+  const actionConfiguration = rule.actions.get(0).configuration;
+  actionConfiguration.put('type', 'application/javascript;version=ECMAScript-2021');
+  actionConfiguration.put('script', '// Code to run when the rule fires:\n' + ruleConfig.execute.toString());
+
   return rule;
 };
 


rules/rules.js
```diff
@@ -280,54 +279,43 @@ const JSRule = function (ruleConfig) {
...
-const registerRule = function (rule) {
-  return currentProvider.addRule(rule);
+/**
+ * Regenerates the rule's source code to show in MainUI by parsing the rule configuration.
+ *
+ * @private
+ * @param {RuleConfig} ruleConfig The rule config describing the rule
+ * @returns {string} the regenerated source code of the rule
+ */
+const generateUIRuleCode = function (ruleConfig) {
```
jpg0, line 300: I am right in saying that this function is attempting to wrap and rewrite script code?

Surely there must be another way to do this. This is typically a very risky thing to do. For example, just at a quick glance, I suspect that this implementation suffers from a lack of proper escaping, so that scripts may contain characters which end the "execute" property of the rule object early. Fixing this escaping may be very difficult.

There will be other problems too - such as mismatching line numbers when failures occur. I would really advise against modifying raw script code, especially in JS. Much better to handle this by providing the needed metadata to the Java layer.

jpg0 (commented):
I have not looked at all the code yet; I have concerns about the approach.

rules/rules.js
florian-h05: > I am right in saying that this function is attempting to wrap and rewrite script code?

No, not really. This function is attempting to regenarate the actual source code of the rule to show it in the UI. But the regenerated code is never executed.

> Surely there must be another way to do this.

I hope so, because I am also not really happy with this implementation. It's just a quick proposal, if it is working that way.

I'm not sure if you've checked out the linked issue, but to summarize: For DSL rules as well as file-based Ruby rules, the roule's source code is shown in the UI, which can be really helpful when you want to know what the rule does.

To solve this in a good way, we'd basically need to access the real source code from the `.js` file, but I've no idea how to accomplish this.

florian-h05: @jpg0 
I‘ve seen that Rules DSL only displays the executed code in MainUI (which makes quite sence, as this is enough to know what the rule is doing). Implementing this for JS is much easier, as we would only need to stringify the callback code and display it as the script in the UI.
Then we add a comment line that says only the execute code is displayer.

WDYT about that?

florian-h05:
Replaced by #199.

PR #175, opened 2022-11-22
[items] ItemHistory: Change return types of min/max between/since to number

florian-h05:
Changes the return types of `ItemHistory.`
- `maximumBetween`
- `maximumSince`
- `minimumBetween`
- `minimumSince`

from string to number.

Possibly breaking.


florian-h05:
@jpg0 Can you review?

jpg0:
> @jpg0 Can you review?

Could you add into the function descriptions what the units of then returned numbers are? I would hope/expect they match the units of the parameters.

florian-h05:
> I would hope/expect they match the units of the parameters.

Not sure what you mean with this.

> what the units of then returned numbers are?

The units would be could of, but I think that the current code wouldn’t even work. I’ll test this later.

I am wondering if we shouldn’t introduce a library for UoM handling like https://github.com/gentooboontoo/js-quantities? WDYT?

jpg0:
Personally I don't believe that we should introduce a units library, at least not yet.

I just wanted to ensure that the return units are easily handled in JS - for example they are milliseconds. I'm not sure what they are from the code/docs.

florian-h05:
> I just wanted to ensure that the return units are easily handled in JS - for example they are milliseconds. I'm not sure what they are from the code/docs.

It seems that:
- for Items from bindings without UoM, just the value is returned, sometimes together with the unit from the Item's state description.
- for Items from bindings with UoM, it seems that the base unit is used (so Wh instead of kWh).

But I'm not sure.

florian-h05:
@jpg0 Are you fine with my little addition to the docs regarding the units?

florian-h05:
@jpg0
I've just noticed that either number or NaN is returned, so either I correct the JSDoc or I modify the function to return number or null as all other functions do.

What would you prefer?

jpg0:
If the other functions return null, then let's keep with that.

Issue #163, opened 2022-10-11
[docs] Add libraries section under advanced topics


rkoshak:
With the addition of `openhab_rules_tools` to openHABian, I think we need to add a libraries section to the docs under Advanced. There we should explain how to properly create a personal library and install it in a way that npm won't delete it when updating packages. Otherwise I imagine some users will become unhappily surprised when their personal library gets wiped out when openHABian runs an update.

We can link to and depend on generic npm tutorials for the most part but I think it's worth the warning because a lot of users will just start editing files under node_modules which is going to cause a problem.

florian-h05:
That’s a good idea.

> install it in a way that npm won't delete it when updating packages. 

Creating a new folder for your personal library inside `node_modules` should work, right?

Editing files inside the folders of npm packages of course isn’t a good idea …

> when their personal library gets wiped out when openHABian runs an update.

openHABian checks for updates and the runs `npm update <package-name>`, so openHABian should only modify the `openhab` and `openhab_rules_tools` folders.

rkoshak:
> Creating a new folder for your personal library inside node_modules should work, right?

Nope. Well, it'll work until you run `npm update` at which point your module won't be in the .json file and `npm` will helpfully remove it.

> openHABian checks for updates and the runs npm update <package-name>, so openHABian should only modify the openhab and openhab_rules_tools folders.

It should but in practice this command too will delete stuff that isn't "installed". Even installing a new library will cause the non-installed modules to be removed.

To get around this, based on the best practices I could figure out, one needs to develop the library outside of the node_modules folder. Then you have some options but the easiest is to tar up your library and use `npm` to install your library from the tar file. That sets everything up so that the personal library is installed and therefore  `npm` won't remove it.

florian-h05:
Okay, thanks for the information.
That‘s definitely something to add to the README.

PR #177, opened 2022-11-25
README: Add a section about libraries

florian-h05:
Fixes https://github.com/openhab/openhab-js/issues/163.

@rkoshak Is that what you‘d expect?

### DIFF
--- a/README.md
+++ b/README.md
@@ -1053,7 +1053,7 @@ var myLibrary = require('my-library');
 
 Note: Only CommonJS `require` is supported, ES module loading using `import` is not supported.
 
-Run the `npm` command from the `automation/js` folder to install third party libraries (CommonJS modules), e.g. from [npm](https://www.npmjs.com/search?q=openhab).
+Run the `npm` command from the `automation/js` folder to install third party libraries, e.g. from [npm](https://www.npmjs.com/search?q=openhab).
 This will create a `node_modules` folder (if it doesn't already exist) and install the library and it's dependencies there.
 
 There are already some openHAB specific libraries available on [npm](https://www.npmjs.com/search?q=openhab), you may also search the forum for details.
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -1040,6 +1041,49 @@ scriptUnloaded = function () {
...
+There are already some openHAB specific libraries available on [npm](https://www.npmjs.com/search?q=openhab), you may also search the forum for details.
+
+#### Creating Your Own Library
+
+You can also create your own personal JavaScript library for openHAB, but you can not just create a folder in `node_modules` and put your library code in it!
+`npm` will remove everything from `node_modules` that has not been properly installed when it is run.
+
+You can follow these steps to create your own library (it's called a Node.js module):
+
+1. Create a separate folder for your library, you may also init a Git repository.
```
sfriedle, line 1066: Maybe it is worth mentioning, that this folder should not ne placed in `automation/js`. Otherwise the engine picks the `*.js` files from the library before they are installed into `node_modules`.

florian-h05: @sfriedle I've added this, thanks.
Can you re-review and say if it looks good to you?

README.md
```diff
@@ -1040,6 +1041,50 @@ scriptUnloaded = function () {
...
 ## Advanced Scripting
 
+### Libraries
+
+#### Third Party Libraries
+
+Loading of third party libraries is supported the same way as loading the openHAB JavaScript library:
+
+```javascript
+var myLibrary = require('my-library');
```
dilyanpalauzov, line 1051: Do ES6 `import` also work?

florian-h05: No, it is not supported.

rkoshak:
Sorry for being too late to review. Last week was a holiday week and I was booked solid with family activities.

I'll have a look and if there is anything I'll open a PR to fix them. I'm sure they are good as is though.

PR #182, opened 2022-11-28
[items] ItemHistory: Return null not NaN from min/max between/since

florian-h05:
@jpg0 
Follow up for https://github.com/openhab/openhab-js/pull/175.


florian-h05:
I’ve verified that it is working properly on my production system.

florian-h05:
@jpg0 Is this good to merge?

PR #183, opened 2022-12-02
[rules] Remove unused rule providers

florian-h05:
Removes methods to select/register another provider for rules. Those are removed, because they go far beyond the use case of most users. As they were not documented, it is not expected that anyone is using them.

@jpg0 Please let me what you think.


jpg0:
LGTM. You have a build error though.

PR #191, opened 2022-12-08
[cache] Reimplementation to use the new core caches

### LINKED ISSUE #184: [cache] Change implementation to the new access-tracking cache in core
With the addition of both a `privateCache` (per script identifier, useful for sharing timers) and a `sharedCache` (shared across all scripts from all languages) in https://github.com/openhab/openhab-core/pull/2887, we have to update our implementation of `cache`.

Backward compatibility must be kept!!

After we have a solution here, we can remove the [SharedCache](https://github.com/openhab/openhab-addons/blob/e979be49114a0e9f234f19ce57b9e11cc2b4493a/bundles/org.openhab.automation.jsscripting/src/main/java/org/openhab/automation/jsscripting/internal/scope/SharedCache.java) from JS Scripting.

My proposal is:
- for the current methods of `cache`:
  - check if the new caches from core are available
  - if they are available, log a deprecation warning and use the shared cache from core instead of the addon‘s shared cache implementation so that we can remove SharedCache from the addon
  - if they are not available (means the library runs on an „old“ openHAB version), do nothing and keep the current behaviour
- add a `private` property to `cache` to allow access to the private cache
- add a `shared` property to `cache` to allow access to the shared cache

@rkoshak @digitaldan @jpg0 WDYT?

florian-h05:
Fixes #184.

Reference https://github.com/openhab/openhab-core/pull/2887.
Reference https://github.com/openhab/openhab-core/pull/3204.

## Description

This implements:
- for the current methods of `cache`:
  - check if the new caches from core are available
  - if they are available, log a deprecation warning and use the shared cache from core instead of the addon‘s shared cache implementation so that we can remove SharedCache from the addon
  - if they are not available (means the library runs on an „old“ openHAB version), do nothing and keep the current behaviour
- add a `private` property to `cache` to allow access to the private cache
- add a `shared` property to `cache` to allow access to the shared cache

Unit test is updated to test the `JSCache` class instead of the exported methods of the `cache` namespace, because testing the exported methods for each case is quite complex and given the fact that the compatibility layer will be removed at some time, I decided to not take that effort.

## Testing

The functionality of `JSCache` is tested by the Jest unit test, the exports are tested by me.


florian-h05:
@digitaldan @jpg0 @rkoshak Can you please review? This is urgent, because we should get this merged before the sixth milestome and the feature freeze comes on Sunday evening.

cache.js
```diff
@@ -1,77 +1,118 @@
...
+  /** @deprecated */
+  get: (key, defaultSupplier) => { return addonSharedJSCache.get(key, defaultSupplier); },
+  /** @deprecated */
+  put: (key, value) => { return addonSharedJSCache.put(key, value); },
+  /** @deprecated */
+  remove: (key) => { return addonSharedJSCache.remove(key); },
+  /** @deprecated */
+  exists: (key) => { return addonSharedJSCache.exists(key); },
+  /**
+   * Private cache for each script.
```
sfriedle, line 101: The comments for `shared` and `private` are swapped.

sfriedle (approved):
LGTM.

cache.js
```diff
@@ -1,77 +1,118 @@
...
- * Associates the specified value with the specified key
- *
- * @memberof cache
- * @param {string} key key with which the specified value is to be associated
- * @param {*} value value to be associated with the specified key
- * @returns {(*|null)} the previous value associated with null, or null if there was no mapping for key
- */
-const put = function (key, value) {
-  return cache.put(key, value);
+const logDeprecationWarning = (funcName) => {
```
rkoshak, line 13: Is there any way we can get the ruleUID or filename to log out as part of this warning? At first some users may get hundreds or these warnings and knowing which rules they are coming from can help in prioritization in dealing with them.

rkoshak (commented):
LGTM assuming either we can't get at the ruleUID or filename or we add that to the log.

cache.js
florian-h05: Thx, I fixed this.

cache.js
florian-h05: By using `console.warn` instead of `log.warn` we are logging to the logger name of the script/rule, so for file-based script to `org.openhab.automation.script.file.filename.js` and for UI to `org.openhab.automation.script.ui.ruleUid`.

florian-h05:
@digitaldan @jpg0 
If you want to give your review, please do it until this evening (+12 hours from now). After this, I'll merge to get this in before the feature freeze; I've already got two reviews so it is not unreviewed.

jpg0:
LGTM!

digitaldan (approved):
LGTM

PR #192, opened 2022-12-09
README: Add note about type defs

florian-h05:
... and mention that Rule Builder currently doesn't support type definitions.

For reference, see https://community.openhab.org/t/require-openhab-rules-is-not-exposing-rule-builder-methods-typescript/141858/3.


hauerm:
Perfectly fine! This will save efforts for others.

PR #195, opened 2022-12-14
Fix broken event object table

wborn:
The table is not properly rendered when reading the documentation in Main UI.

I created this PR to prevent it from being copy/pasted into the add-ons repo which has the same issue see https://github.com/openhab/openhab-addons/pull/13952.


florian-h05 (approved):
LGTM, thanks!

Issue #197, opened 2022-12-18
`toString()` doesn‘t properly represent Items, Things


florian-h05:
Calling the `toString()` method on native Java openHAB Items returns a string that represents the Item with it‘s metadata etc. (e.g. `F2_Office_Main_Light (Type=SwitchItem, State=OFF, Label=Deckenlampe AZ, Category=light, Tags=[Lightbulb], Groups=[LichterAZ])`), but calling `toString()` on the JS version tries to stringify the object: 

```
{
  "rawItem": {},

  "history": {

    "rawItem": {}

  },

  "semantics": {

    "rawItem": {}

  }

} 
```

[`ThingImpl`](https://github.com/openhab/openhab-core/blob/main/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/internal/ThingImpl.java) doesn‘t provide a `toString` method, maybe one should be introduced?

FYI have a look at the `toString` method of [`GenericItem`](https://github.com/openhab/openhab-core/blob/97299c7dcb3c0527071629903cf1b839aad08a37/bundles/org.openhab.core/src/main/java/org/openhab/core/items/GenericItem.java#L273).

This issue is just a reminder for myself and to track this.

@stefan-hoehn Das hier ist das Problem wieso das Loggen eines Items von Blockly aus auf JS Scripting nicht mehr so funktioniert wie auf Nashorn.

stefan-hoehn:
Thx, @florian-h05 , I appreciate if that would be done. Do you need support here?

florian-h05:
Thanks for asking, I’d need a review for my PR https://github.com/openhab/openhab-js/pull/198.

florian-h05:
Closing as #198 has been merged and I don't see the need in providing a Thing serialization from core.

PR #198, opened 2022-12-18
[items & things] Add `toString` overrides & Clean-Ups

florian-h05:
Reference #197 (fixes the JS library side).

## To Dos

- [x] Regenerate type definitions
- [x] Update CHANGELOG


florian-h05:
@jpg0 Can I remove thing builder and channel builder here?

jpg0:
> @jpg0 Can I remove thing builder and channel builder here?

Are they used? I don't mind whether they go or stay, I suppose it depends on the purpose of the library. I used them to define and register things and channels in scripts (and then refer to them from other scripts). I did this to move all my configuration into JS, and then could reference everything via JS imports instead of name strings.

However if this library is limited to rules based on existing definitions, then this functionality should be removed.

Maybe @digitaldan has an opinion?

digitaldan:
> Maybe @digitaldan has an opinion?

I would error on the side of simplicity. Since it's undocumented and probably unused by most, i think removing it makes sense.

florian-h05:
So we agree that we remove that?

Can one of you please review here (although the changes apart from the removal are minimal).

florian-h05:
@jpg0 Can you please review here? You can ignore the `*.d.ts` files, these type definitions are auto generated.

digitaldan (approved):
LGTM!

PR #199, opened 2022-12-20
[rules] Display `execute` code of `JSRule` in MainUI

florian-h05:
Replaces/Superseds #173.

Fixes https://github.com/openhab/openhab-addons/issues/12929.

## Description

With this PR, the `JSRule` API adds additional information to the rule‘s action, so that MainUI is able to display the code that is executed when the rule runs together with a comment note.
The `execute` function is stringified by calling the JS `toString` method on the function, Rule Builder is currently not supported (some useless code is displayed).

## To Do

- [ ] Update CHANGELOG

@jpg0 WDYT?


jpg0:
I'm no typescript expert, but the plain JS looks good to me. Happy to merge if there is no danger in the ts?

florian-h05:
The ts code is no danger, it can never break the library's functionality. The worst thing that can happen with the ts files is, that the autocompletion doesn't work. I have just updated the CHANGELOG, feel free to merge @jpg0!

PR #201, opened 2022-12-22
Add `js-quantities` for units of measurement handling

### LINKED ISSUE #102: Add utilities to improve QuantityType handling

### LINKED ISSUE #200: Introduce `js-quantities` for units of measurement handling
I propose to add the [js-quantities](https://www.npmjs.com/package/js-quantities) library to openhab-js, because:

- We need a good unit of measurement handling in the JS layer to work with UoM returned from the Java layer
- Blockly would heavily profit from js-quantities, as we can have something like: 1 W +  1 kW and it calculates 1001 W
- It seems to be a much requested feature in the community (I can't provide links, but I guess others can)
- js-quantities has **NO** dependencies and is only 30 kB bundles, so we don't have to loose anything if we add it
- js-quantities covers many units (if you know all of them, respect), see https://github.com/gentooboontoo/js-quantities/blob/master/src/quantities/definitions.js

/cc @digitaldan @jpg0 @rkoshak @stefan-hoehn @ghys

florian-h05:
Replaced by #206.

Fixes #102.
Closes #200.
Fixes https://github.com/openhab/openhab-webui/issues/836.

## Description

This PR adds the fantastic [js-quantities](https://www.npmjs.com/package/js-quantities) library to provide easy `QuantityType` handling and unit conversions.

Given the [openHAB UoM Docs](https://next.openhab.org/docs/concepts/units-of-measurement.html#list-of-units), I have created unit tests to check compatibility of all base units from openHAB with the js-quantities library.

After polyfilling support for some additional unit notations (e.g. 'Ws' is `W*s`) and four units (see https://github.com/florian-h05/openhab-js/blob/js-quantities/quantities.js, the required math is linked), the new `Quantity` namespace is able to handle all units from openHAB except octet for data amount and mired for Temperature.

FYI: The addition of this increases the bundle site from 857KB to 932KB, which is minimal given the advantage of having the library and it's functionality. js-quantities has **no** dependencies.

I will add the docs in a follow-up PR.

## Testing

The additional units I've added are tested by unit tests as well to verify that they work mathematically correct.

/ping @stefan-hoehn @rkoshak @digitaldan 

@digitaldan @jpg0  I would like to get your opinion on the addition of such a library and possibly also some review, there is not much to review. The real code is just 60 lines.

### DIFF
--- a/index.js
+++ b/index.js
@@ -29,5 +29,6 @@ module.exports = {
   get utils () { return require('./utils'); },
   get osgi () { return require('./osgi'); },
   get cache () { return require('./cache'); },
-  get time () { return require('./time'); }
+  get time () { return require('./time'); },
+  get Quantity () { return require('js-quantities'); }
 };
--- a/index.js
+++ b/index.js
@@ -30,5 +30,5 @@ module.exports = {
   get osgi () { return require('./osgi'); },
   get cache () { return require('./cache'); },
   get time () { return require('./time'); },
-  get Quantity () { return require('js-quantities'); }
+  get Quantity () { return require('./quantities'); }
 };
--- a/.gitignore
+++ b/.gitignore
@@ -1,9 +1,10 @@
 .DS_Store
 .vscode
 .idea
+coverage
 dist
 docs
 node_modules
 openhab-*.tgz
 out
-coverage
+types/quantities.*
--- a/package-lock.json
+++ b/package-lock.json
@@ -11,10 +11,12 @@
             "dependencies": {
                 "@js-joda/core": "^4.3.1",
                 "@js-joda/timezone": "^2.11.1",
+                "js-quantities": "^1.7.6",
                 "parse-duration": "^0.1.1"
             },
             "devDependencies": {
                 "@types/jest": "^28.1.8",
+                "@types/js-quantities": "^1.6.3",
                 "acorn": "^8.6.0",
                 "docdash": "^1.2.0",
                 "jest": "^28.1.3",
@@ -1674,6 +1676,12 @@
                 "expect": "^28.0.0",
                 "pretty-format": "^28.0.0"
             }
+        },
+        "node_modules/@types/js-quantities": {
+            "version": "1.6.3",
+            "resolved": "https://registry.npmjs.org/@types/js-quantities/-/js-quantities-1.6.3.tgz",
+            "integrity": "sha512-YwBhHz9Ad7GH/hvg8P9qoKoqG5+xYsM2AQpb96jCDS9HbIj3m24RPgoY2xKPUd5LYAIkaL26igRQQrETf4Gwxg==",
+            "dev": true
         },
         "node_modules/@types/json-schema": {
             "version": "7.0.9",
@@ -6344,6 +6352,11 @@
             "engines": {
                 "node": ">=8"
             }
+        },
+        "node_modules/js-quantities": {
+            "version": "1.7.6",
+            "resolved": "https://registry.npmjs.org/js-quantities/-/js-quantities-1.7.6.tgz",
+            "integrity": "sha512-h6TH1xK1u/zdjD26M6kKVthZONJSDTIRzrohbqOILfJAyanWHGlJLWuAWkSMtqi8k/IxshStsc97Pkf8SL9yvA=="
         },
         "node_modules/js-tokens": {
             "version": "4.0.0",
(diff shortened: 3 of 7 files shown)


florian-h05:
@digitaldan Thanks for your positive feedback in the issue #200. 

Can you please have a look at the code (the actual important code is just 60 lines: `quantities.js`)? 
FYI: I have provided links to all the math but the units test check whether the math is correct.
Please add your approval if you are happy, I will merge then.

The major part of changes is just package-lock and unit tests.

jpg0:
@florian-h05 I think that it's a great addition to incorporate unit handling. The main question I have is why it's not possible to use the native openHAB logic to do this? Presumably it also needs to handling parsing strings into unit representations? It seems weird to have the core handle this, but also include additional logic in a JS library to do the same thing (or similar, which isn't great). It also means that other scripting languages don't benefit.

I'm guessing that there are fundamental reasons why this cannot be handled in core, but was interested to know why?

index.js
```diff
@@ -29,5 +29,6 @@ module.exports = {
   get utils () { return require('./utils'); },
   get osgi () { return require('./osgi'); },
   get cache () { return require('./cache'); },
-  get time () { return require('./time'); }
+  get time () { return require('./time'); },
+  get Quantity () { return require('./quantities'); }
```
digitaldan, line 33: Should 'Quantity' be 'quantity' here to match the other exports which are lowercase?

digitaldan (changes requested):
Looks good, thanks!  Seems like a pretty minimal inclusion, so not a whole lot to review, which is always a good thing.  The other comment i have is i think there needs to be a section in the readme describing its use and a couple of common examples (temp and power especially) .

index.js
florian-h05: I have also asked that myself, but given that the `Quantity` export is a factory function itself and the other exports aren't, I decided to make it uppercase:

```javascript
const { items, things, rules, Quantitiy } = require('openhab');

items.getItem('Kitchen_Light');
things.getThing('lightbulb')
rules.JSRule(...)

// But

Quantitiy('1.364 nm')
```

florian-h05:
> there needs to be a section in the readme describing its use and a couple of common examples (temp and power especially) 

Yes, this is something for a follow-up PR.

@digitaldan Have you read @jpg0 comments on the issue #200? What do you think regarding his question?

florian-h05:
@digitaldan I'll also add some conversion logic to the addon like you did for JS-Joda so that our JS `Quantity` is translated to a `QuantityType` when "send" to an openHAB API.

index.js
digitaldan: ahh, makes sense.

digitaldan (approved):
besides the small merge conflict, LGTM !

florian-h05:
@digitaldan Note that this has been replaced by #206. I‘ll close this here!

PR #205, opened 2022-12-24
[items] ItemHistory: Add `previousStateTimestamp` method

florian-h05:
Required for https://github.com/openhab/openhab-webui/issues/1597.

## To Do

- [x] Update CHANGELOG
- [x] Update README

/cc @stefan-hoehn


items/item-history.js
```diff
@@ -1,7 +1,5 @@
-// const { ZonedDateTime } = require('@js-joda/core');
-const time = require('../time');
+const time = require('@js-joda/core');
```
digitaldan, line 1: out of curiosity, whats the benefit of not using the internal time class that also has this import ? (so now there are multiple places importing an external lib)

florian-h05: When importing the internal time library, I was not able to get the type definitions for it and then the functions that return a ZonedDateTime had no type definitions.

sfriedle: You can import `ZonedDateTime` from @js-joda/core via a `@typedef` statement in the docs. Then the `require` can point to our own `time.js`:

```js
const time = require('../time');

/**
 * @typedef { import('@js-joda/core').ZonedDateTime } ZonedDateTime
 */
```

In the JSDoc comments `ZonedDateTime` can be referenced directly, e. g.: 

```js
  /**
   * Gets the average value of the state of a given Item between two certain points in time.
   *
   * @param {(ZonedDateTime | Date)} begin begin
   * @param {(ZonedDateTime | Date)} end end
   * @param {string} [serviceId] Optional persistence service ID, if omitted, the default persistence service will be used.
   * @returns {(number | null)}
   */
  averageBetween (begin, end, serviceId) {
    return this._decimalOrNull(PersistenceExtensions.averageBetween(this.rawItem, ...arguments));
  }
```

florian-h05: @sfriedle Thanks for the suggestion, I‘ve already tried this a few times and it always caused probems with the JSDoc generation, but luckily I found a JSDoc plugin that solves that issue.

florian-h05:
@digitaldan I am now using the internal import (as before), I was able to fix the problem with typedefs & JSDoc. This is ready for merging now.

digitaldan (approved):
LGTM !

PR #206, opened 2022-12-25
[Quantity] Initial contribution: Support UoM handling by wrapping QuantityType

### LINKED ISSUE #102: Add utilities to improve QuantityType handling

### LINKED ISSUE #200: Introduce `js-quantities` for units of measurement handling

florian-h05:
Replaces #201.
Fixes #102.
Closes #200.
Fixes https://github.com/openhab/openhab-webui/issues/836.

## Description

This implementation of a Units of Measurement/Quantity handling functionality wraps the QuantityType from openHAB instead of introducing a new library.

I’ll add the docs later if this one gets accepted.

## To Do

- [x] Update CHANGELOG

@digitaldan @jpg0 
Can you please review this one?

The `*.d.ts` files can be ignored, they are autogenerated by the typescript compiler.


digitaldan (approved):
This looks great, and i think the fact that it will match up with the official openHAB quantity api and docs is a very good thing.

PR #209, opened 2022-12-27
Fix & Improve TS type declarations

florian-h05:
@sfriedle I've rebased your branch and added a few improvements. Can you please have a look at my changes?

## Description

This PR improves the type definitions by solving bug and extending them:

* Java types moved to global scope and can be referenced from anywhere
* Missing type imports were added to several `js` files.
* A type testing script was added to check the type definitions for problems
* Type definitions for Rule Builder were enabled, but require further improvements. This will be done in another PR.

Also-by: Florian Hotze <florianh_dev@icloud.com>
Signed-off-by: Stefan Friedle <stefan@friedle.de>


sfriedle (commented):
@florian-h05 you are faster then me :-)

I added one comment to `actions.js`, the rest LGTM. The generated type declarations can be processed by the TypeScript compiler.

But the typings at whole … I think there are some quirks or issues, if one uses the lib with TypeScript. I'm just at the beginning to migrate my personal rules to openhab-js (using TypeScript), but I stumble over some things, the TypeScript compiler rejects or my IDE (VS Code) can't resolve.

florian-h05:
@sfriedle Every help with the type defs is welcome and appreciated.

PR #211, opened 2022-12-28
[Quantity] Rename comparison methods & Add docs

florian-h05:
Follow-up to #206.

## Description

- Add Quantity documentation to README


florian-h05:
@rkoshak Can you please have a look at this?

README.md
```diff
@@ -853,6 +854,75 @@ var timestamp = time.ZonedDateTime.now().plusMinutes(5);
 console.log(timestamp.getMillisFromNow());
 ```
 
+### Quantity
+
+The `Quantity` class greatly simplifies Units of Measurement (UoM)/Quantity handling by providing unit conversion, comparisons and mathematical operations.
+
```
stefan-hoehn, line 860: I would define what is meant by a Quantity ("the measurement and the unit of measurement) and the unit itself.

README.md
```diff
@@ -853,6 +854,75 @@ var timestamp = time.ZonedDateTime.now().plusMinutes(5);
...
+Internally using the openHAB `QuantityType`, which relies on [`javax.measure`](https://unitsofmeasurement.github.io/unit-api/), it supports all units and dimensions that openHAB supports - see [Units of Measurement documentation](https://www.openhab.org/docs/concepts/units-of-measurement.html#list-of-units) for an (incomplete) overview.
+If you unit is not listed in the UoM docs, it is very likely that it is still supported, e.g. the Angstrom Å for very small lengths (1 Å = 10 nm).
+
+#### Creation
+
+`Quantity(value)` can be used as a factory (without `new` - don't use it as a constructor), pass an amount **and** a unit to it to create a new `Quantity` object:
+
```
stefan-hoehn, line 867: Please explain how the actual needs to be transformed, i.e. Nm -> N*m, m²/s² -> m^2/s^2, -> whitespace not allowed as I think it is not quite obvious at least that a * has to be used in Nm.

I would recommend to take some "special" examples from here https://www.openhab.org/docs/concepts/units-of-measurement.html#list-of-units and show how they would be transformed (eg. ɡₙ, °dH, W/m², µW/cm²)

README.md
```diff
@@ -853,6 +854,75 @@ var timestamp = time.ZonedDateTime.now().plusMinutes(5);
...
+qty = Quantity('1'); // unitless quantity
+
+// Not allowed:
+qty = Quantity('m');
+qty = Quantity('1 N m'); // * is required
+qty = Quantity('1 m^2 s^2'); // / is required
+qty = Quantity(1 m2/s2); // ^ is required
+```
+
+`value` can be a string, a `Quantity` instance or an openHAB Java [`QuantityType`](https://www.openhab.org/javadoc/latest/org/openhab/core/library/types/quantitytype).
```
stefan-hoehn, line 885: The argument `value` ...

README.md
```diff
@@ -853,6 +854,75 @@ var timestamp = time.ZonedDateTime.now().plusMinutes(5);
...
+
+`value` strings have the `$amount $unit` format and must follow these rules:
+
+- `$amount` is required and a number as string
```
stefan-hoehn, line 889: `$amount` is required with a number provided as a string

README.md
```diff
@@ -853,6 +854,75 @@ var timestamp = time.ZonedDateTime.now().plusMinutes(5);
...
+- `$unit` is optional (unitless quantities are possible) and can have a prefix like `m` (milli) or `M` (mega)
+- If a `$unit` contains `*`, `/` and `^`, they are not optional and must be present.
+
+#### Conversion
+
+It is possible to convert a `Quantity` to another unit or get the value as integer or float:
```
stefan-hoehn, line 895: It is possible to convert a `Quantity` to new `Quantity` with a different unit

README.md
```diff
@@ -853,6 +854,75 @@ var timestamp = time.ZonedDateTime.now().plusMinutes(5);
...
+
+`Quantity` provides the following methods for comparison:
+
+- `equal(value)` ⇒ `boolean`: this `Quantity` === another `Quantity`
+- `largerThan` ⇒ `boolean`: this `Quantity` > another `Quantity`
+- `largerThanOrEqual` ⇒ `boolean`: this `Quantity` >= another `Quantity`
+- `smallerThan` ⇒ `boolean`: this `Quantity` < another `Quantity`
+- `smallerThanOrEqual` ⇒ `boolean`: this `Quantity` =< another `Quantity`
+
+`value` can be a string or another `Quantity`, for the string the same rules apply as described above.
```
stefan-hoehn, line 917: the same rules apply for the string  as described above

stefan-hoehn, line 917: value` can be a string or a `Quantity`
another -> a or did you intentionally use "another"

README.md
```diff
@@ -853,6 +854,75 @@ var timestamp = time.ZonedDateTime.now().plusMinutes(5);
...
+
+#### Mathematical Operators
+
+- `add(value)` ⇒ `Quantity`: `value` can be a string or another `Quantity`, for the string the same rules apply as described above
```
stefan-hoehn, line 921: another -> a (next lines as well) or did you intentionally use "another"

stefan-hoehn (changes requested):

README.md
rkoshak: Maybe provide a link to the UoM page where you say "Units of Measurement (UoM)". I see you linked to it below, but I like the idea of moving that link up to here.

README.md
rkoshak: I'm cautious about the "(incomplete)" part. It's incomplete now but will it be forever? Once it becomes complete, will we remember to come and remove that here? 

I'd leave it out and create an issue to go make the linked to page complete.

Doesn't the link to the javax.measure list *all* the units? So while our docs only show the most commonly used ones in a home automation context, they can get to the complete list from there, right?

> `Quantity(value)` can be used as a factory (without `new` - don't use it as a constructor)

For someone who doesn't know programming, this sentence is confusing and scary. I'd switch it around and put the "without `new` outside the parenthetical and put "factory" inside.

>`Quantity(value)` is used without `new` (it's a factory, not a construcor).

I left out the "can" because as I understand it, this is a *must*, not a *can*.

README.md
```diff
@@ -865,6 +854,89 @@ var timestamp = time.ZonedDateTime.now().plusMinutes(5);
...
+```
+
+`toUnit` returns a new Quantity with the given unit or `null`, if conversion to that unit is not possible.
+
+#### Comparison
+
+`Quantity` provides the following methods for comparison:
+
+- `equal(value)` ⇒ `boolean`: this `Quantity` === another `Quantity`
+- `largerThan` ⇒ `boolean`: this `Quantity` > another `Quantity`
```
rkoshak, line 912: We usually say "greaterThat" instead of "largerThan". Both make sense but "greater" is closer to how we are taught in elementary school. I'm not sure if that's just a US thing or English speaking world universal.

README.md
```diff
@@ -865,6 +854,89 @@ var timestamp = time.ZonedDateTime.now().plusMinutes(5);
...
+var floatValue = qty.float;
+```
+
+`toUnit` returns a new Quantity with the given unit or `null`, if conversion to that unit is not possible.
+
+#### Comparison
+
+`Quantity` provides the following methods for comparison:
+
+- `equal(value)` ⇒ `boolean`: this `Quantity` === another `Quantity`
```
rkoshak, line 911: I know there is a subtle difference between `==` and `===`. Do we need a footnote or anything or will this work as expected for users who don't know the difference?

rkoshak (commented):
Sorry for the delay. I agree with all @stefan-hoehn 's comments and added just a few of my own.

README.md
florian-h05: > Doesn't the link to the javax.measure list all the units? So while our docs only show the most commonly used ones in a home automation context, they can get to the complete list from there, right?

Unfortunately no -- I haven't found a list of the units supported by javax.measure. Even not in the JavaDoc.

README.md
florian-h05: I guess I should remove the `=` signs here.
You cannot compare the quantities by using the equal sign `=`. The `equal` method checks, whether two quantities are equal in the sense of 500 cm is equal to 5 m.

README.md
florian-h05: https://www.w3schools.com/js/js_comparisons.asp is using greater and less, not larger and smaller.

I have no strong opinion here, I guess we can change this although it is released, but I don't expect many to use it as it was released a few hours ago and is undocumented.

README.md
florian-h05: > show how they would be transformed (eg. ɡₙ, °dH, W/m², µW/cm²)

Actutally they work all except the standard gravity, I have no idea how to get that work.

README.md
rkoshak: I'd leave it up to you. It does throw a little bit of head tilt when seeing it as larger instead of greater.

README.md
florian-h05: I will see if I can change it.

README.md
florian-h05: I will check if I can change it.

florian-h05:
@rkoshak @stefan-hoehn Can you please have a look at the current version?

README.md
stefan-hoehn: > We usually say "greaterThat" instead of "largerThan"

Really "greaterTha**t**"?

stefan-hoehn:
LGTM now

README.md
florian-h05: I’d change it to greaterThan

This is how w3schools has it and the Mozilla WebDev Docs also have it with than, see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators#relational_operators.

florian-h05:
@stefan-hoehn 
Let’s wait for Rich and see if he’s happy with the changes.

README.md
rkoshak: > Really "greaterThat"?

No, that's a typo. It's greaterThan and lessThan, same as in the w3schools link @florian-h05 posted.

florian-h05: @rkoshak 
Since the current version already has 60 downloads from npm, I’ll publish the changed method names together with the changes from #212 in a new major version to follow semver.

rkoshak:
LGTM. Thanks! I'm looking forward to having this support!

florian-h05:
It is actually already included in the current 3.2.0 library release, but it’s not documented there. 
I will try to publish a new version with the changed method names and the refactored metadata API soon, but before I want to do some other improvements.

Anyway, I wish you all a good new year (that begins at different times for us) 🎆

PR #218, opened 2023-01-01
Update bundling mechanism for globals injection caching

florian-h05:
This contains the required changes for the openhab-js injection caching inside the JS Scripting addon in https://github.com/openhab/openhab-addons/pull/14135.

## Description

- Add injection script
- Modify webpack to bundle the library with the injection script as entrypoint
- Set webpack mode to production (now the bundle size is only about 450 KB, before it was about 900 KB)

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,8 +1,10 @@
 # WARNING: Please keep in mind that npm disregards .gitignore when there is a .npmignore file!
 
+@openhab-globals.js
 .github
-.vscode
 .idea
+.markdownlint.yaml
+.vscode
 coverage
 DEPLOY.md
 doc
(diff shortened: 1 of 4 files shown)


digitaldan (approved):
LGTM.  FYI, we will probably need to also update DEPLOY.md as well if we change the output name.

florian-h05:
Oh, thanks for that reminded.

We also need a new release for this - I feel it is a bit to early to publish a 4.0 (due to some breaking changes it will be a new major). WDYT about a 3.2.1 release?

florian-h05:
I decided to create a 3.2.x branch out of the 3.2.0 tag and cherrypick this on top of it.

digitaldan:
>  WDYT about a 3.2.1 release?

Sounds good to me.

PR #222, opened 2023-01-05
[items] metadata: return configuration as js object

stefan-hoehn:
Signed-off-by: Stefan Höhn <mail@stefanhoehn.com>

return the configuration as js object for easier processing of its attributes


florian-h05 (approved):
LGTM, thanks!

PR #223, opened 2023-01-05
[items] Extend `metadata` & `itemchannellink` APIs

### LINKED ISSUE #220: [items] Extend `metadata` & `itemchannellink` APIs
See https://github.com/openhab/openhab-js/pull/212

I've been working with creating, updating , and deleting items with procedural scripts, and it would be much more powerful if sometimes I didn't have to specify item names or the precise metadata values etc. 
Sometimes I would like to delete all channel links from all items, or a selection of items, or a single item. 
Sometimes I would like to delete all invalid channel links.

florian-h05:
Closes #220.


items/metadata/metadata.js
```diff
@@ -74,9 +74,45 @@ const removeMetadata = function (itemOrName, namespace) {
   };
 };
 
+// TODO: Move implementation to openHAB Core
```
florian-h05, line 77: @digitaldan Do you think we should move the underlying implementation of this method to openHAB core?

FY: I decided to use a JS Map instead of an object because IMO Maps are easier to document both in JSDoc and TypeScript than "dynamic" objects. I guess only advanced users will use this method, but it is well documented so Map shouldn't be a problem here.

florian-h05:
@digitaldan 
This is ready for review now.
Please don't merge - I need to update type defs and CHANGELOG before.

digitaldan:
Sounds good, I have just headed out for the day , will take a look when I'm back , thanks !

items/metadata/metadata.js
digitaldan: >Do you think we should move the underlying implementation of this method to openHAB core? 

Hmm,  if we think this can be reused by other scripting engines, then maybe.  I don't do much with metadata in my own rules, so i don't have a ton of familiarity with querying/manipulating item metadata

items/metadata/metadata.js
```diff
@@ -74,9 +74,45 @@ const removeMetadata = function (itemOrName, namespace) {
...
+ * @example
+ * var meta = items.metadata.getItemMetadata('Kitchen_Light');
+ * var namespaces = Array.from(meta.keys());
+ * var stateDescription = meta.get('stateDescription');
+ *
+ * @memberof items.metadata
+ * @param {string} itemName the name of the Item
+ * @returns {Map<string, {configuration: object, value: string}>} Map of metadata
+ */
+const getItemMetadata = function (itemName) {
```
digitaldan, line 90: So we have `item.getMetaData(namespace)`, `metadata.getMetadata(itemName, nameSpace)`, and `metadata. getItemMetadata(itemName)`.  I'm not entirely clear on the the purpose of the two latter functions?  They sound very similar.

florian-h05: `Item.getMetadate(namespace)` and `items.metadata.getMetadata(itemOrName, namespace)` are the same functions. The prior is just a "shortcut" on the Item object so you don't need to call the latter method with the Item or Item name.

`getItemMetadata(itemOrName)` however gets ALL metadata an Item has, not only one metadata namespace like the two functions above.

items/metadata/metadata.js
florian-h05: > i don't have a ton of familiarity with querying/manipulating item metadata

Same on my side. This PR is more on user request than on personal interest.

Core already has a method to get the namespaces of all metadata on an Item, but this method doesn't deliver the metadata itself.

But I could imagine other script enginges want to use something like this as well.
I would propose to leave the implementation here to get this finished soon, and then I will see if/when I can open a core PR.

items/metadata/metadata.js
digitaldan: Right, so i'm wondering if we should try and combine `items.metadata.getMetadata` and `metadata. getItemMetadata` into single function, maybe make the namespace optional?  Does that complicate the return structure too much ?

florian-h05: Sounds like a great idea, we could switch to return an object in case we want to get all metadata. 
I’m just not sure yet how to document the return, but I guess this should be possible using by using some JSDoc „magic“. But type definitions/auto-completion will get less exact I guess.

florian-h05: I was able to combine the get and remove metdata methods each into a single one that accepts the namespace as an optional parameter. Return type documentation seems to be no real problem, I have checked the generated type defs, they look good as well.

florian-h05:
@digitaldan This is ready for review now.
Please don't merge, type defs and CHANGELOG updates are missing.
The checks are failing unrelated.

items/metadata/metadata.js
digitaldan: Thanks.  I see now, so it returns
```js
{
 ns: {
 value: v,
 configuration: c
},
...more ns entries ...
}
```
or
```js
{
 value: v,
 configuration: c
}
```

Depending on parameter.  I can see the need to for two methods ( i was not sure of the return structure before).   I wonder then if we should keep two methods and call them  `getAllMetadata()` and `getMetadata(namespace)` (or something similar, i'm not great at naming things).  This is where function overloading comes in handy in Java, but i guess in JS dynamic return types are probably equivalent ?    I'm really good either way, wdyt?

florian-h05: I put some work in the documentation of the return type in JSDoc and the type defs and I think keeping these "unified" methods therefore is okay :+1: 

How is your review going?

digitaldan (approved):
LGTM

items/metadata/metadata.js
digitaldan: Good, on a plane with spotty internet :-)

florian-h05: Ooh. You don't need to be on a plane for spotty internet ;-) 
At my home it is often too much for the internet when it's evening and everyone wants to stream something ...

PR #228, opened 2023-01-09
[items] Return HistoricItem from persistence where applicable

### LINKED ISSUE #226: [items] Persistence wrapper discards timestamps
In one of my rules, I use `PersistenceExtensions` to grab a value with timestamp to fill 2 items with it:
```
var PersistenceExtensions = Java.type("org.openhab.core.persistence.extensions.PersistenceExtensions");
var ZonedDateTime = Java.type("java.time.ZonedDateTime");
var LocalTime = Java.type("java.time.LocalTime");
var windSpeedItem = items.getItem("GardenWindSensorGustSpeed").rawItem;
var maxToday = PersistenceExtensions.maximumSince(windSpeedItem, ZonedDateTime.now().with(LocalTime.MIDNIGHT));
if (maxToday) {
  items.getItem("GardenWindSensorVirtualWindSpeedMaxToday").postUpdate(maxToday.getState().toString());
  items.getItem("GardenWindSensorVirtualWindSpeedMaxTodayTime").postUpdate(time.ZonedDateTime.parse(maxToday.getTimestamp().toString()));
}
```

Since I need the timestamp, I can't use `item.history`. It would be great if the `ItemHistory` object could handle that situation, e.g. by returning an object containing both state and timestamp (the latter as `ZonedDateTime`), for all applicable calls:
- `maximumSince`
- `maximumBetween`
- `minimumSince`
- `minimumBetween`
- `previousState` (currently implemented as 2 methods, only case where timestamp can be returned)

That way, the above rule would become much shorter:
```
var midnight = time.toZDT().toStartOfDay(); // see #225 ;-)
var maxToday = items.getItem("GardenWindSensorGustSpeed").history.maximumSince(midnight);
if (maxToday) {
  items.getItem("GardenWindSensorVirtualWindSpeedMaxToday").postUpdate(maxToday.state);
  items.getItem("GardenWindSensorVirtualWindSpeedMaxTodayTime").postUpdate(maxToday.timestamp);
}
```

maniac103:
For minimumSince, minimumBetween, maximumSince and maximumBetween the caller might care about the timestamp of the returned state, so return an object which contains both state and timestamp. Additionally, remove previousStateTimestamp and make previousState return an HistoricItem instead for API symmetry.

Fixes #226


maniac103:
I'll submit a PR to update the tern definitions once this is accepted. Open question: shall I update the `CHANGELOG.md` file?

items/item-history.js
```diff
@@ -395,22 +376,29 @@ class ItemHistory {
...
    */
-  _decimalOrNull (result) {
-    return result === null ? null : result.toBigDecimal();
+  _historicItemOrNull (result) {
+    if (result === null) {
+      return null;
+    }
+    return {
+      state: result.getState().toString(),
+      rawState: result.getState(),
```
maniac103, line 399: What I'm a bit unsure about is that maximum*/minimum* used to return a number, which the new code does not return anymore, since previousState can essentially be anything, not only a number. AFAICT, there's two options:

- leave it as is (string as convenience + Java state for special needs, which can also e.g. be forwarded to postUpdate())
- add a decimalState (or something) field, which becomes NaN if the state is not a number

Thoughts?

florian-h05: I see the problem, that we cannot generally return `state` as a number.
Given the fact that minimum*/maximum* did return a string for a long time and the switch to number was not so long ago, I‘d consider it okay to leave is as it currently is.

maniac103: Is there a way to check whether `result.getState()` is a `DecimalType` in the JS code? If so, I'd simply fill `decimalState` (or `numberState`?) only in that case.

florian-h05: Yes, that’s possible.

You need to import the `DecimalType` and perform an `instanceof` check, but given the fact that `QuantityType` and `PercentType` can contain decimal values as well I wonder if it is not easier to perform a `isNaN` on the string state instead of checking the type.

types/items/item-history.d.ts
```diff
@@ -1,4 +1,23 @@
 export = ItemHistory;
+
+/**
+ * Item state with timestamp
+ */
+export type HistoricItem = {
```
florian-h05, line 6: Instead of directly defining the `HistoricItem` type, rather defined it in JSDoc at the top of the item-history.js file (after the imports), because JSDoc needs some type definition as well and unfortunantely does not read the .d.ts files.

```javascript
/**
 * @typedef {object} HistoricItem
 * @property {string} state Item state
 * @property {HostState} rawState Raw Java state
 * @property {time.ZonedDateTime} timestamp timestamp of historic Item
 */
```

items/item-history.js
```diff
@@ -395,22 +376,29 @@ class ItemHistory {
   /**
    * @private
    */
-  _timestampOrNull (result) {
```
florian-h05, line 398: You can remove the `_stateOrNull` method above.

florian-h05 (changes requested):
Code LGTM, just some comments "around the code".

items/item-history.js
```diff
@@ -237,25 +237,21 @@ class ItemHistory {
    * @param {(time.ZonedDateTime | Date)} begin begin
    * @param {(time.ZonedDateTime | Date)} end end
    * @param {string} [serviceId] Optional persistence service ID, if omitted, the default persistence service will be used.
-   * @returns {(number | null)} state or null
+   * @returns {(HistoricItem | null)} historic item or null
```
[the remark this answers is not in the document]
maniac103: But it's not an `Item` that's returned here. I tried to follow the Javadoc of the actual `HistoricItem`.

items/item-history.js
maniac103: I think `QuantityType` and `PercentType` both are subclasses of `DecimalType`, so the instanceof check should work for all 3?

types/items/item-history.d.ts
maniac103: I don't fully understand yet. Do you mean your suggested change *replaces* the item-history.d.ts one, or should this be added *in addition*?

items/item-history.js
florian-h05: I am not sure how the instanceof check behaves for Java classes, you can test by constructing a new QuantityTyoe after importing it and then performing the instanceof check.

items/item-history.js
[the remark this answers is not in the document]
florian-h05: The JavaDoc often use lowercase, but newer docs have Item uppercase. I know that’s not an actual Item, but I’d prefer to follow the docs.

types/items/item-history.d.ts
florian-h05: Have you written the HistoricItem in the `item-history.d.ts` file yourself?

items/item-history.js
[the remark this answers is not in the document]
maniac103: The way we spell this in the Android app is 'Item' when referring to the OH concept 'item' and 'item' when referring to some other 'thing', and I think this also is the general style of the docs. That's what I'd follow here as well, TBH...

types/items/item-history.d.ts
maniac103: Yes, that's why it's in the diff ;-)

items/item-history.js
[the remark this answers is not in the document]
florian-h05: I agree that we don’t have a real openHAB Item here, but some type of.
We can also leave it that way, that’s nothing big to discuss about.

items/item-history.js
maniac103: I went the easier route and checked for `isNaN`.

types/items/item-history.d.ts
maniac103: Done.

maniac103:
>> Open question: shall I update the CHANGELOG.md file?

> Yes, please.

Done.

README.md
```diff
@@ -490,6 +490,19 @@ var item = items.getItem('KitchenDimmer');
 console.log('KitchenDimmer averageSince', item.history.averageSince(yesterday));
 ```
 
+The `HistoricItem` object contains the the Item state in various representations and a timestamp.
```
florian-h05, line 493:
```suggestion
The `HistoricItem` object contains the Item state in various representations and a timestamp.
```

Please also add a short paragraph/sentence about the `timestamp` property.

florian-h05 (approved):
LGTM, apart from one doc recommendation.

items/item-history.js
```diff
@@ -388,29 +377,32 @@ class ItemHistory {
...
+  _historicItemOrNull (result) {
+    if (result === null) {
+      return null;
+    }
+    const rawState = result.getState();
+    const numericState = parseFloat(rawState.toString());
+    return {
+      state: rawState.toString(),
+      rawState: rawState,
+      numericState: numericState == NaN ? null : numericState,
```
florian-h05, line 403:
```suggestion
      numericState: isNaN(numericState) ? null : numericState,
```

To fix the linting errors.

maniac103: Silly mistake on my part :-/ ... fixed.

florian-h05 (approved):
LGTM, thanks!

PR #229, opened 2023-01-13
fix executeCommandLine without a Duration

rkoshak:
Without a `Duration` argument, `executeCommandLine` always returns `null`. I removed the example that appears to show calling it without a `Duration` and getting the result of the echo.  That can't happen as it's currently implemented.

Also, we need to use `time.Duration` right? I'm pretty sure that's the case. If I'm wrong and that's not the case, we should show importing java.time.Duration in the example.

I think I forgot to sign off on this. This is really minor doc update, I hope the exception can apply. If not I'll figure out how to fix it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -582,16 +582,14 @@ See [openhab-js : actions.Exec](https://openhab.github.io/openhab-js/actions.htm
 Execute a command line.
 
 ```javascript
-
 // Execute command line.
 actions.Exec.executeCommandLine('echo', 'Hello World!');
 
 // Execute command line with timeout.
-var Duration = Java.type('java.time.Duration');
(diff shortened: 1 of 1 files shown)


florian-h05:
I‘ve just checked the core JavaDoc, `executeCommandLine` without a Duration is only `void`, no wonder it is always `null`.

> Also, we need to use time.Duration right?

Yes (we don‘t need, Java Duration also works but is not the nice way).

> This is really minor doc update, I hope the exception can apply.

IMO the exception applies here.

florian-h05 (approved):
LGTM, thanks!

I‘ve fixed something I noticed while reviewing this.

PR #233, opened 2023-02-04
[items] Allow Item lookup by name directly on the `items` namespace

### LINKED ISSUE #232: Add sendCommand and postUpdate to `items`
This continues an idea from https://github.com/openhab/openhab-webui/pull/1664#issuecomment-1413941287

# The Problem

Compared to jRuby and Rules DSL, JS Scripting can be *really* verbose. While this isn't necessarily a problem by itself, it does become irksome when it comes to commanding and updating Items, largely because when all you have is the name of the Item you must pull the `Item` Object from `items` first.

Rules DSL:

    sendCommand('ItemName', 'ON')

jRuby:

    items["ItemName"].on
    items["ItemName"].command ON.
    items["ItemName"] >> ON

JS Scripting

    items.getItem('ItemName').sendCommand('ON');
    actions.BusEvent.sendCommand('itemName', 'ON');

This wouldn't be so irksome if sending a command or updating an Item you only have the name for is a rare occurrence. But in my case it's not. It's the most common case for me. And often I have to do it more than once in the same rule.

# A Proposed Solution

This is by no means the only way to do this but my first thought is to add `sendCommand('itemorname', 'string')` and `postUpdate('itemorname', 'string)` to `items`.

    items.sendCommand('itemName', 'ON');

I'm not tied to `itemorname`. I can see arguments for why we would want to only support `name` instead of the Item Object. But I like the idea that it would work with both. 

# Another Idea

Another idea which I've no idea if it's possible or if possible if it's an antipattern would be if `items` could behave like a dict too.

    items['ItemName'].sendCommand('ON');

But that feels weird for it to be both a dict and an Object with functions and something tells me if this were feasible it would have been done this way to begin with. But if this makes sense, it would help make a number of other use cases a little more terse and easier to read.

The good news is either proposal would be non-breaking.

digitaldan:
Fixes #232


digitaldan:
Hmm, i need to update the docs as well, i do that shortly

digitaldan:
Ok, was not sure the best way to document in JSDoc, this was the best i could come up with.

florian-h05 (approved):
LGTM, thanks!

Just a few minor things I will take care of:
- static Item lookup throws exception if Item is missing instead of returning `null` as intended
- README: Add comments to remaining items.getItem calls
- type def updates
- CHANGELOG update

digitaldan:
Thanks!

PR #234, opened 2023-02-04
[items] Add `numericState` & `quantityState` to `Item` & `HistoricItem`

### LINKED ISSUE #231: Add methods to Item to get  the state as Quantity or Number
Opening the issue so I don't forget and, if someone gets to it before me that's good too.

`HistoricItem` has a `numericState` method to return the state already cast/converted to a value we can do math with.  This would be very useful to have on the `Item` class too.

In addition, this method should be `Quantity` aware or we should also provide an alternative `quantityState` method to make it easier to deal with the new `Quantity` API.

Or maybe it makes sense to bring back the whole `getStateAs()` discussion which can work for this as well as stuff like getting a Dimmer Item's ON/OFF state and the like.

florian-h05:
Fixes #231.

This adds two new properties to the Item class:

- `numericState`: the Item state as float or null if not a number
- `quantityState`: the Item state as JS Quantity or null if the state is not Quantity-compatible

The JS `HistoricItem` is extended with the `quantityState` as well.

@digitaldan Please don't merge, type def and CHANGELOG updates are missing.


items/items.js
```diff
@@ -89,39 +91,61 @@ class Item {
...
   }
 
   /**
-   * The raw state of the Item, as a Java {@link https://www.openhab.org/javadoc/latest/org/openhab/core/types/state State object}.
+   * Numeric representation of Item state, or `null` if state is not numeric
+   * @type {number|null}
+   */
+  get numericState () {
+    const numericState = parseFloat(this.rawState.toString());
+    return isNaN(numericState) ? null : numericState;
```
digitaldan, line 131: Should we return null, or NaN ?  Since this is a native javascript return type, i'm a little on the fence.

digitaldan (approved):
LGTM!
I had one small question/comment, but am fine either way you decide

items/items.js
florian-h05: I have no strong opinion here. 
If we return NaN, we should change the unreleased `HistoricItem` as well.

What kind of speaks against returning NaN here is, that quantityState returns null if not possible so returning null here as well would make it more inline.

florian-h05:
@digitaldan This is ready for merging now. 
I will leave the merging to you so you can answer to my answer above.

items/items.js
digitaldan: I think null works just as well then.

PR #243, opened 2023-02-14
[items] Fix error handling for `quantityState`

florian-h05:
HistoricItem creation failed for Items that were not Quantity-compatible because the catch mechanism for QuantityError had a design issue.

I have verified that this fix is working.


stefan-hoehn:
I checked the fix and it is working correctly for me. The error has gone away.

florian-h05:
@digitaldan Can you please review and merge? The other PR as well, they are both relatively short.

items/item-history.js
```diff
@@ -408,8 +409,12 @@ class ItemHistory {
     try {
       quantityState = Quantity(rawState.toString());
     } catch (e) {
-      if (e instanceof QuantityError) quantityState = null;
-      throw Error('Failed to create "quantityState": ' + e);
+      if (e instanceof QuantityError) {
+        log.debug(`Failed to create "quantityState" for Item ${this.name}:`, e);
```
digitaldan, line 413: Do we want to leave debug messages like this here?  I wonder if this is just adding noise?

digitaldan (approved):
I just had one comment about the debug messages,  I personally would error on being silent, unless its something like a depreciation warning which requires user intervention (and would be removed eventually)

Other then that , LGTM

items/item-history.js
florian-h05: I agree, I will remove this logging.

PR #244, opened 2023-02-14
[utils] Add access to library version

### LINKED ISSUE #238: Add function that returns version of library
As I've been playing with and incorporating the changes in the 4.0 release it occurs to me that there are a few places where it would be handy to know which version of the openhab-js library is installed. I'm mostly thinking about rule template situations where it might have completely different approaches based on the base library.

Obviously I can infer the version based on whether certain functions and properties exist or not but that feels brittle.

My idea was to have a function, probably in `utils` that returns the version of openhab.js. Initially I thought it could just load the version from package.json but I'm not certain that file would be available to the library when using the version that comes with the add-on. If there isn't another way short of needing to remember to change some constant in the code every time a release is made this may not be worth doing. 

It's probably a half baked idea but surely this is something that Node developers have had to solve, right?

florian-h05:
Fixes #238.

This PR adds a constant to the `utils` namespace that included the current version string.
I had some concerns that this would not work with webpack, but I have tested this and it works as both webpacked version and npm package.


florian-h05:
Please don‘t merge, I missed to update the type defs.

digitaldan (approved):
LGTM!

rkoshak:
Just to show why I asked for this, see https://github.com/rkoshak/openhab-rules-tools/blob/main/helpers.js#L140.

My rewritten rule templates now have at the top

    helpers.validateLibraries('4.1.0', '2.0.1');

If either the installed openhab-js or my openhab_rules_tools are too old for the template to run it bails with a meaningful exception. I'm on a big kick to make sure all my rule templates generate meaningful errors and this is a big help. Thanks!

PR #259, opened 2023-03-29
[Quantity] Fix `symbol` availability & handling in `sendCommand`/`postUpdate`

florian-h05:
See invividual commit messages.


florian-h05:
@digitaldan Can you please leave merging to me and only review? I want to do a rebase merge to keep the two individual commits.

items/items.js
```diff
@@ -52,6 +51,27 @@ const managedItemProvider = osgi.getService('org.openhab.core.items.ManagedItemP
...
+ *
+ * Numbers and strings are passed.
+ * Objects should implement `toOpenHabString` (prioritized) or `toString` to return a openHAB compatible representation.
+ *
+ * @private
+ * @param {*} value
+ * @returns {*}
+ */
+const _toOpenhabString = (value) => {
+  if (typeof value === 'number' || typeof value === 'string') {
```
digitaldan, line 65: Since the function returns value by default do you need to check for number or string here?

README.md
```diff
@@ -383,9 +383,9 @@ Calling `getItem(...)` or `...` returns an `Item` object with the following prop
   - .getMetadata(namespace) ⇒ `object|null`
   - .replaceMetadata(namespace, value, configuration) ⇒ `object`
   - .removeMetadata(namespace) ⇒ `object|null`
-  - .sendCommand(value): `value` can be a string or a [`time.ZonedDateTime`](#time)
-  - .sendCommandIfDifferent(value) ⇒ `boolean`: `value` can be a string or a [`time.ZonedDateTime`](#time)
-  - .postUpdate(value): `value` can be a string or a [`time.ZonedDateTime`](#time)
+  - .sendCommand(value): `value` can be a string, a [`time.ZonedDateTime`](#time) or a [`Quantity`](#quantity)
```
digitaldan, line 386: Does value need to be a string ?  Or do we take the string representation of the value (and do proper conversion of ZonedDateTime & Quantity)?

items/items.js
florian-h05: If we want to pass though number and don’t call `toString` on numbers, we need to check at least for number here.
`sendCommand` and `postUpdate` accept both number, string and openHAB types (see https://www.openhab.org/javadoc/latest/org/openhab/core/model/script/actions/busevent). 

I think converting a number to string can lead to loss of precision, I also wonder if it would be better to pass JS Quantity as QuantityType instead of its string representation. WDYT?

README.md
florian-h05: We can take number, string and openHAB types.
openhab-js types like Quantity or JS-Joda ZonedDateTime need to converted.
However I am not sure what you mean with your question, sorry.

items/items.js
digitaldan: > If we want to pass though number and don’t call toString on numbers,

Gotcha , i missed that `value.toString === 'function'` would match a number

> I also wonder if it would be better to pass JS Quantity as QuantityType instead of its string representation. WDYT?

I so rarely use UoM that i'm cautious to give an opinion one way or another.

README.md
digitaldan: Nevermind ;-) I was not sure if when finally calling the send event method in the @runtime package  if the value is eventually converted into a string if it does not match a standard type (like a number, or something the addon specifically registers to convert  )

digitaldan (approved):

items/items.js
florian-h05: Let's leave as it is and merge :+1:

PR #261, opened 2023-04-24
[items] Fix Item semantics `location` & `equipment`

florian-h05:
Reported in the community, see https://community.openhab.org/t/itemsemantics/146294.

Fixes an issue where the `location` and `equipment` properties were not available because `getItem` failed.
This happened because `items.js` was imported from `item-semantics.js` and `items.js` imported `item-semantics.js`, which lead to a circular require and failed.
This hacky fix passes the `getItem` method via the `ItemSemantics` constructor.

/cc @rkoshak


digitaldan:
> This hacky fix passes the getItem method via the ItemSemantics constructor.

Yeah, a little hacky ;-)  but i get the alternative might be a larger refactoring, maybe keep a refactor it in mind if this comes up again ?   I think  its ok for now.

digitaldan (approved):
LGTM

florian-h05:
> maybe keep a refactor it in mind if this comes up again ? I think its ok for now.

Yeah, if this comes up once again I'd refactor it into the items file, this is the only other solution.

PR #262, opened 2023-04-24
Use `function` instead of `const = function`

florian-h05:
... and enable ESLint rule to enforce this code style. 
Additionaly, use arrow functions for callbacks.

This has some benefits:
- Functions don't have to be ordered in their call order (`const = function` is only available when code execution reaches its line)
- `function` keyword clearly indicates a function in contrast to `const = function` when skimming over code
- Better stack trace since arguments are named and not anonymous


digitaldan:
Hmm, yeah i have flipped flopped between the two styles, mainly b/c some code base tended to use one or the other.  I tend to agree i like declaring functions with 'function',  if  for no other reason then it looks nicer to me ;-)   

Quick question on prefixing private functions with `_` since i see that a lot in the diffs (not related to your changes) , its something i have always though of as hacky, especially since there's no enforcement of it in the language, its just a pattern that has emerged over time.  While i have actually never seen a code base use it, ES 2019 i believe introduced actual private functions by prefixing them with '#',  i would assume grallvm supports this?  Might be a option for us? 

In any case since there a lot of changes in the PR, i just did a quick look over and trust its been well tested.

digitaldan (approved):

florian-h05:
@digitaldan FYI I just rebased this PR and enabled the ESLint rule to enforce this code style.

> While i have actually never seen a code base use it, ES 2019 i believe introduced actual private functions by prefixing them with '#', i would assume grallvm supports this? Might be a option for us?

AFAIK the `#` is only available for private fiels and methods of classes, not function declarations. I've tested this syntax some time ago on openHAB 3.4.x and GraalJS in ECMA2021 mode there does not support this syntax. Note that the functions declared are private if they are not exported with `module.exports`, the `_` is just a naming pattern. In openhab-js, no `_private` functions are exported except when unit tests need them.

> In any case since there a lot of changes in the PR, i just did a quick look over and trust its been well tested.

It's actually not that much changes (most changes probably come from the automatically regenerated type defs), but I am testing this in my production setup (as with nearly any PR).

digitaldan:
>FAIK the # is only available for private fiels and methods of classes

Right  🤦

PR #263, opened 2023-04-30
[Quantity] Minor improvements & Make `quantityState` `null` when unit is missing

florian-h05:
* Improve availability of `symbol` property (once again) by extracting it from the string representation of the Quantity instead of getting it from the Java layer.
* Log a warning when unit conversion fails (especially important for Blockly users).
* Return `null` from `quantityState` if Quantity would be unit-less (without unit) since there is no advantage over using a plain number. This is especially important for the UoM blocks in Blockly.


stefan-hoehn:
supports [#1864](https://github.com/openhab/openhab-webui/pull/1864)
Tested:
- Log a warning when unit conversion fails works as expected
- Return null from quantityState if Quantity would be unit-less ...

Both work as expected

digitaldan (approved):
LGTM

PR #264, opened 2023-05-13
Clarification: Nashorn will not be availabe from openHAB 4 on

dilyanpalauzov:
This is my understanding.


florian-h05:
... then your understanding is wrong. 

The README of openhab-js is already up to date for openHAB 4: Nashorn will be/is removed from core and therefore will not be available by default. But it is available as the addon that the docs mention here.

This library was never and is not compatible with Nashorn.

PR #271, opened 2023-06-11
[rules] Preserve disabled state of switchable rules after restarting

simonihmig:
When a `SwitchableJSRule` has been switched off and I change something in my script so that it restarts, then after recreation of the rules the switch item would correctly be in `OFF` state given its historical state, but the rule it represents would still be enabled. Which basically means whenever your script restarts, all previously disabled rules would get enabled, without noticing...

This should fix it, tested locally.

### DIFF
--- a/rules/rules.js
+++ b/rules/rules.js
@@ -324,6 +324,7 @@ function SwitchableJSRule (ruleConfig) {
       item.sendCommand('ON');
     }
   }
+
   RuleManager.setEnabled(rule.getUID(), item.state !== 'OFF');
 }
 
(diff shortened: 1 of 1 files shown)


florian-h05 (approved):
LGTM, thanks!

rules/rules.js
```diff
@@ -324,6 +324,8 @@ function SwitchableJSRule (ruleConfig) {
       item.sendCommand('ON');
     }
   }
+
```
florian-h05, line 327:
```suggestion

```

This fixes the lint error.

PR #292, opened 2023-08-21
Ignore more IDE generated files with .gitignore

wborn:
These unwanted files are also often generated when creating projects in Eclipse and IntelliJ IDEA.


florian-h05 (approved):
LGTM, thanks!

PR #291, opened 2023-08-13
[rules] Rule Builder: Add DateTime & TimeOfDay triggers & Improve type defs

florian-h05:
- Improve type defs (hide private fields), closes #217.
- Add DateTime trigger, closes #280.
- Add TimeOfDay trigger.
- Add debug logging to trigger creation.

PR #293, opened 2023-08-22
Fix example in documentation

jlaur:
Fix Rule Builder example.


jlaur:
I just by coincidence looked at #291. If I understand it correctly, the README is generated from those files, where you already fixed this example?

florian-h05:
I have fixed this example in the openhab-js README in #291.
The addon README is in most parts just a copy of the openhab-js README.

jlaur:
> I have fixed this example in the openhab-js README

Right, now saw it was indeed the README, I thought I was looking at .js files. 😊

PR #307, opened 2023-11-09
Add timezone ID when parsing 8601 String, fix typo

rkoshak:
# Add ZoneId to ISO8601 parsed strings to toToday works over DST changeovers

# Description

During DST changes, any ISO8601 String that is parsed by `time.toZDT()`, including times from a DateTime Item's state, lack the ZoneId. However, it is this ZoneId which tells Joda when DST changes occur so when `toToday()` is called, the time is not adjusted to account for DST. This addition adds the system default ZoneId to the parsed ZonedDateTime.

### DIFF
--- a/time.js
+++ b/time.js
@@ -90,13 +90,13 @@ function _parseISO8601 (isoStr) {
     LOCAL_DATE: /^(-?(?:[1-9][0-9]*)?[0-9]{4})-(1[0-2]|0[1-9])-(3[01]|0[1-9]|[12][0-9])$/, // YYYY-MM-DD
     LOCAL_TIME: /^\d{2}:\d{2}(:\d{2})?(\.\d+)?$/, // hh:mm or hh:mm:ss or hh:mm:ss.f
     LOCAL_DATE_TIME: /^(-?(?:[1-9][0-9]*)?[0-9]{4})-(1[0-2]|0[1-9])-(3[01]|0[1-9]|[12][0-9])T\d{2}:\d{2}(:\d{2})?(\.\d+)?$/, // LOCAL_DATE and LOCAL_TIME connected with "T"
-    ISO_8160_FULL: /^(-?(?:[1-9][0-9]*)?[0-9]{4})-(1[0-2]|0[1-9])-(3[01]|0[1-9]|[12][0-9])T(2[0-3]|[01][0-9]):([0-5][0-9])(:[0-5][0-9])?(\.\d+)?(Z|[+-]\d{2}(:\d{2})?)/
+    ISO_8601_FULL: /^(-?(?:[1-9][0-9]*)?[0-9]{4})-(1[0-2]|0[1-9])-(3[01]|0[1-9]|[12][0-9])T(2[0-3]|[01][0-9]):([0-5][0-9])(:[0-5][0-9])?(\.\d+)?(Z|[+-]\d{2}(:\d{2})?)/
   };
   switch (true) {
     case REGEX.LOCAL_DATE.test(isoStr): return time.ZonedDateTime.of(time.LocalDate.parse(isoStr), time.LocalTime.MIDNIGHT, time.ZoneId.SYSTEM);
     case REGEX.LOCAL_TIME.test(isoStr): return time.ZonedDateTime.of(time.LocalDate.now(), time.LocalTime.parse(isoStr), time.ZoneId.SYSTEM);
     case REGEX.LOCAL_DATE_TIME.test(isoStr): return time.ZonedDateTime.of(time.LocalDateTime.parse(isoStr), time.ZoneId.SYSTEM);
-    case REGEX.ISO_8160_FULL.test(isoStr): return time.ZonedDateTime.parse(isoStr);
+    case REGEX.ISO_8601_FULL.test(isoStr): return time.ZonedDateTime.parse(isoStr).withZoneSameLocal(time.ZoneId.SYSTEM);
   }
   return null;
 }


florian-h05 (approved):
LGTM, thanks!

time.js
```diff
@@ -90,13 +90,13 @@ function _parseISO8601 (isoStr) {
...
     LOCAL_DATE_TIME: /^(-?(?:[1-9][0-9]*)?[0-9]{4})-(1[0-2]|0[1-9])-(3[01]|0[1-9]|[12][0-9])T\d{2}:\d{2}(:\d{2})?(\.\d+)?$/, // LOCAL_DATE and LOCAL_TIME connected with "T"
-    ISO_8160_FULL: /^(-?(?:[1-9][0-9]*)?[0-9]{4})-(1[0-2]|0[1-9])-(3[01]|0[1-9]|[12][0-9])T(2[0-3]|[01][0-9]):([0-5][0-9])(:[0-5][0-9])?(\.\d+)?(Z|[+-]\d{2}(:\d{2})?)/
+    ISO_8601_FULL: /^(-?(?:[1-9][0-9]*)?[0-9]{4})-(1[0-2]|0[1-9])-(3[01]|0[1-9]|[12][0-9])T(2[0-3]|[01][0-9]):([0-5][0-9])(:[0-5][0-9])?(\.\d+)?(Z|[+-]\d{2}(:\d{2})?)/
   };
   switch (true) {
     case REGEX.LOCAL_DATE.test(isoStr): return time.ZonedDateTime.of(time.LocalDate.parse(isoStr), time.LocalTime.MIDNIGHT, time.ZoneId.SYSTEM);
     case REGEX.LOCAL_TIME.test(isoStr): return time.ZonedDateTime.of(time.LocalDate.now(), time.LocalTime.parse(isoStr), time.ZoneId.SYSTEM);
     case REGEX.LOCAL_DATE_TIME.test(isoStr): return time.ZonedDateTime.of(time.LocalDateTime.parse(isoStr), time.ZoneId.SYSTEM);
-    case REGEX.ISO_8160_FULL.test(isoStr): return time.ZonedDateTime.parse(isoStr);
+    case REGEX.ISO_8601_FULL.test(isoStr): return time.ZonedDateTime.parse(isoStr).withZoneSameLocal(time.ZoneId.SYSTEM);
```
florian-h05, line 99: One thing that just came to my mind before merging, but after approving:

What if a user passes an ISO String containing a timezone? In this case his passed in timezone would be overwritten.

We’d have to only set the zone when no zone is passed in. So either we make two regexes out of the single ISO8601 and differ between with and without timezone, or we append a `[SYSTEM]` of the string does not include a timezone.

florian-h05: This REGEX should only match ISO8601 with timezone in square brackets:

`^(-?(?:[1-9][0-9]*)?[0-9]{4})-(1[0-2]|0[1-9])-(3[01]|0[1-9]|[12][0-9])T(2[0-3]|[01][0-9]):([0-5][0-9])(:[0-5][0-9])?(\.\d+)?(Z|[+-]\d{2}(:\d{2})?\[[\w/]*\])`

And this one should only match ISO8601 without timezone in square brackets:

`^(-?(?:[1-9][0-9]*)?[0-9]{4})-(1[0-2]|0[1-9])-(3[01]|0[1-9]|[12][0-9])T(2[0-3]|[01][0-9]):([0-5][0-9])(:[0-5][0-9])?(\.\d+)?(Z|[+-]\d{2}(:\d{2})$)`

FYI: I usually try this out on https://regex101.com

rkoshak: Eeek,  that's right. We probably don't want to just replace the timezone ID if it's used in the String, only when it's not there should we assume the system default.

> We’d have to only set the zone when no zone is passed in. So either we make two regexes out of the single ISO8601 and differ between with and without timezone, or we append a [SYSTEM] of the string does not include a timezone.

There's a third option which could make the regex simpler. We can parse the ISO8601 string as is. Then we can get the zone ID from the new ZDT. If that ID is `[+-]\d{2}(:\d{2}` we need to set the zone id to the system default. 

I don't know if that makes a difference from a maintainability perspective.

florian-h05: IMO having two different regexes is more straight forward, than checking the zone id of the ZDT.

rkoshak: Added a second regex and it works for Strings.

However, it still fails to work for Items and rawState. See below.

rkoshak:
We have fixed the problem but only for Strings. 😭  I should have thought of this.

It does not work when the DateTime Item or the raw DateTimeType is passed to `toZDT()`.

See the following code with the results:

```
var now = time.toZDT();
var beforeDSTstr = time.toZDT(items.TestDateTime.state);
var beforeDSTdtt = time.toZDT(items.TestDateTime.rawState);
var beforeDSTdti = time.toZDT(items.TestDateTime);
console.info('Now:              ' + now);
console.info('Before DST Str:   ' + beforeDSTstr);
console.info('To Today Str:     ' + beforeDSTstr.toToday());
console.info('Before DST DTT:   ' + beforeDSTdtt);
console.info('To Today DTT:     ' + beforeDSTdtt.toToday());
console.info('Before DST DTI:   ' + beforeDSTdti);
console.info('To Today DTI:     ' + beforeDSTdti.toToday());
```

Results:

```
2023-11-14 14:59:42.409 [INFO ] [nhab.automation.script.ui.scratchpad] - Now:              2023-11-14T14:59:42.405-07:00[SYSTEM]
2023-11-14 14:59:42.410 [INFO ] [nhab.automation.script.ui.scratchpad] - Before DST Str:   2023-11-03T14:02:04.274-06:00[SYSTEM]
2023-11-14 14:59:42.411 [INFO ] [nhab.automation.script.ui.scratchpad] - To Today Str:     2023-11-14T14:02:04.274-07:00[SYSTEM]
2023-11-14 14:59:42.412 [INFO ] [nhab.automation.script.ui.scratchpad] - Before DST DTT:   2023-11-03T14:02:04.274-06:00
2023-11-14 14:59:42.413 [INFO ] [nhab.automation.script.ui.scratchpad] - To Today DTT:     2023-11-14T14:02:04.274-06:00
2023-11-14 14:59:42.413 [INFO ] [nhab.automation.script.ui.scratchpad] - Before DST DTI:   2023-11-03T14:02:04.274-06:00
2023-11-14 14:59:42.414 [INFO ] [nhab.automation.script.ui.scratchpad] - To Today DTI:     2023-11-14T14:02:04.274-06:00
```

I was so focused on the toString that I forgot about Items and states too. 

`toToday()` should work on ZDS produced from these too, right? I'm not sure the best way to handle this. Should I add the fix in both of those cases in `untils.javaZDTToJsZDT`? If I do that do we even need both REGEX anymore? Do we need to sit back and think about this some more?

time.js
```diff
@@ -134,7 +136,7 @@ function _parseString (str) {
   // ISO8601 Time, Date, or DateTime string
   try {
     // Blockly compatibility with user input without the "T"
-    const newStr = /^(-?(?:[1-9][0-9]*)?[0-9]{4})-(1[0-2]|0[1-9])-(3[01]|0[1-9]|[12][0-9]) \d{2}:\d{2}/.test(str) ? str.replace(' ', 'T') : str;
+    const newStr = /^(-?(?:[1-9][0-9]*)?[0-9]{4})-(1[0-2]|0[1-9])-(3[01]|0[1-9]|[12][0-9]) \d{2}:\d{2}?\[[\w/]*\])*\]?/.test(str) ? str.replace(' ', 'T') : str;
```
florian-h05, line 139: Our CI action is not happy with your regex, because it seems you have a bracket too much or are missing a bracket.
However that does not matter, because we don‘t need to adjust the regex.
`RexExp.prototype.test()` returns true if there is a match, and the existing regex matches the beginning of all valid inputs (see https://github.com/openhab/openhab-webui/blob/dfbdcf23b34afb1b352c19d88a2cbbb309b5e941/bundles/org.openhab.ui/web/src/assets/definitions/blockly/blocks-dateoffsets.js#L221).

time.js
florian-h05: See my comment https://github.com/openhab/openhab-js/pull/307#discussion_r1395564442.

florian-h05:
> Should I add the fix in both of those cases in untils.javaZDTToJsZDT? If I do that do we even need both REGEX anymore? Do we need to sit back and think about this some more?

Yes, we need both RegEx(es) anymore, because IMO it should be possible to parse a ZDT from an (ISO) String containing a timeline or not containing a tz. 
Theoretically, the utils function could be replaced by converting the Java ZDT to string and  parsing it using the parseISO function, but string parsing has very bad performance. I did some performance benchmarking when developing the utils function,

I am against changing the existing utils function, because IMO the utils should  simply convert and not set any defaults. I would add a warning regarding timezones to the utils method’s docs and introduce a private method to time.js to default to the system time zone for all usages of the utils method inside time.js

florian-h05:
> introduce a private method to time.js to default to the system time zone

Or should we add this as a new utils method?

rkoshak:
> Yes, we need both RegEx(es) anymore, because IMO it should be possible to parse a ZDT from an (ISO) String containing a timeline or not containing a tz.

That was working before though, wasn't it? The only problem was that those Strings that were parsed without the zone id wouldn't handle `toToday()` correctly on the day of a DST change. 

>  I would add a warning regarding timezones to the utils method’s docs and introduce a private method to time.js to default to the system time zone for all usages of the utils method inside time.js

That seems reasonable.

> Or should we add this as a new utils method?

I like the idea of keeping like stuff together. I'm not certain why the utils method is in utils in the first place nor where that util method is used. This feels like a similar function though so it makes sense to me to make a new utils function that time.js calls. 

I'll fix the regex so the builds become happier.

florian-h05:
> That was working before though, wasn't it? The only problem was that those Strings that were parsed without the zone id wouldn't handle toToday() correctly on the day of a DST change.

Exactly. Those RegExp can be kept as they currently are in this PR.

> I'm not certain why the utils method is in utils in the first place nor where that util method is used.

When helping out @jlaur with JS scripting examples for his energy prices binding, we noticed that there might be the need to have such a method. Of course toZDT is able to handle that as well, but given the number of javaToJS functions we already had in utils, I thought it would be useful to provide a simple conversions for ZDT there as well. If I want to convert a Java type to a JS type, I always look first at utils, as this is the dedicated place for such stuff.

jlaur:
> When helping out @jlaur with JS scripting examples for his energy prices binding, we noticed that there might be the need to have such a method. Of course toZDT is able to handle that as well, but given the number of javaToJS functions we already had in utils, I thought it would be useful to provide a simple conversions for ZDT there as well. If I want to convert a Java type to a JS type, I always look first at utils, as this is the dedicated place for such stuff.

For reference:
- https://github.com/openhab/openhab-addons/pull/14376#discussion_r1189121993
- https://github.com/openhab/openhab-addons/pull/14376#discussion_r1197040988

rkoshak:
That makes sense. Because this is very much like the existing utils method I think it should probably go there and then have the time method call the new function from utils instead of the old one. 

It will probably be late today but stay tuned for a new commit. The changes are super easy thankfully.

florian-h05:
@rkoshak Whats the status here? Should I take over?

rkoshak:
Unfortunately I've not been able to get back to this. It's still at the top of my todo list but stuff like mitigating a burst pipe and planned family activities have eaten up most of the time I have to work this. 

If you have the time I'm happy to let you take over.

PR #322, opened 2024-02-05
README: Add JS-Joda locale installation and usage

### LINKED ISSUE #54: time.ZonedDateTime.parse and format fails with "use locale" error

rkoshak:
Closes #54.

Adds a section that explains when one might need to install a JS-Joda locale, why, and how to install a locale, import and use it in a rule.

### DIFF
--- a/README.md
+++ b/README.md
@@ -833,7 +833,7 @@ See [JS-Joda](https://js-joda.github.io/js-joda/) for more examples and complete
 
 Occasionally, one will need to parse a non-supported date time string or generate one from a ZonedDateTime.
 To do this you will use [JS-Joda DateTimeFormatter and potentially your Locale](https://js-joda.github.io/js-joda/manual/formatting.html).
-However, all the locales combine to a unacceptibly large size to ship them with the openhab-js library.
+However, shipping all the locales with the openhab-js library would lead to an unacceptable large size.
 Therefore if you attempt to use the `DateTimeFormatter` and receive an error saying it cannot find your locale, you will need to manually install your locale and import it into your rule.
 
 [JS-Joda Locales](https://github.com/js-joda/js-joda/tree/master/packages/locale#use-prebuilt-locale-packages) includes a list of all the supported locales.
--- a/README.md
+++ b/README.md
@@ -829,6 +829,23 @@ actions.Exec.executeCommandLine(time.Duration.ofSeconds(20), 'echo', 'Hello Worl
 
 See [JS-Joda](https://js-joda.github.io/js-joda/) for more examples and complete API usage.
 
+### Parsing and Formatting
+Occasionally, one will need to parse a non-supported date time string or generate one from a ZonedDateTime.
+To do this you will use [JS-Joda DateTimeFormatter and potentially your Locale](https://js-joda.github.io/js-joda/manual/formatting.html).
+However, all the locales combine to a unacceptibly large size to ship them with the openhab-js library.
+Therefore if you attempt to use the `DateTimeFormatter` and receive an error saying it cannot find your locale, you will need to install your locale and import it into your rule.
+
+[JS-Joda Locales](https://github.com/js-joda/js-joda/tree/master/packages/locale#use-prebuilt-locale-packages) includes a list of all the supported locales.
+Each loacle consists of a two letter language indicator followed by a "-" and a two letter dialect indicator: e.g. "EN-US".
+Installing a locale can be done through the command `npm install @js-joda/locale_de-de` from the *$OPENHAB_CONF/automation/js* folder.
+
+To import and use a local into your rule you need to require it and create a `DateTimeFormatter` that uses it.
+
+```javascript
+var Locale = require('@js-joda/locale_de-de').Locale.GERMAN;
+var formatter = time.DateTimeFormatter.ofPattern('MM/dd/yy HH:mm').withLocale(Locale);
+```
+
 #### `time.toZDT()`
 
 There will be times when this automatic conversion is not available (for example when working with date times within a rule).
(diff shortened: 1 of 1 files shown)


rkoshak:
Somehow I think I ended up making this PR to my own fork and not here where it belongs. Trying again since I see no evidence that my PR exists anywhere but on my own fork.

florian-h05:
This PR is a the correct place. 
Can you please fix the markdown lint?

rkoshak:
I think I messed up the DCO on the markdown linting fix. It was just inserting a newline so hopefully I can squeak under and not have to redo it. The original commit was properly signed off.

florian-h05:
I’ll set DCO to pass.

README.md
```diff
@@ -829,6 +829,24 @@ actions.Exec.executeCommandLine(time.Duration.ofSeconds(20), 'echo', 'Hello Worl
 
 See [JS-Joda](https://js-joda.github.io/js-joda/) for more examples and complete API usage.
 
+### Parsing and Formatting
```
florian-h05, line 832:
```suggestion
#### Parsing and Formatting
```

README.md
```diff
@@ -829,6 +829,24 @@ actions.Exec.executeCommandLine(time.Duration.ofSeconds(20), 'echo', 'Hello Worl
...
+
+Occasionally, one will need to parse a non-supported date time string or generate one from a ZonedDateTime.
+To do this you will use [JS-Joda DateTimeFormatter and potentially your Locale](https://js-joda.github.io/js-joda/manual/formatting.html).
+However, all the locales combine to a unacceptibly large size to ship them with the openhab-js library.
```
florian-h05, line 836:
```suggestion
However, shipping all the locales with the openhab-js library would lead to an unacceptable large size of it.
```

README.md
```diff
@@ -829,6 +829,24 @@ actions.Exec.executeCommandLine(time.Duration.ofSeconds(20), 'echo', 'Hello Worl
...
+Therefore if you attempt to use the `DateTimeFormatter` and receive an error saying it cannot find your locale, you will need to install your locale and import it into your rule.
```
florian-h05, line 837:
```suggestion
Therefore if you attempt to use the `DateTimeFormatter` and receive an error saying it cannot find your locale, you will need to manually install your locale and import it into your rule.
```

README.md
```diff
@@ -829,6 +829,24 @@ actions.Exec.executeCommandLine(time.Duration.ofSeconds(20), 'echo', 'Hello Worl
...
+
+[JS-Joda Locales](https://github.com/js-joda/js-joda/tree/master/packages/locale#use-prebuilt-locale-packages) includes a list of all the supported locales.
+Each loacle consists of a two letter language indicator followed by a "-" and a two letter dialect indicator: e.g. "EN-US".
+Installing a locale can be done through the command `npm install @js-joda/locale_de-de` from the *$OPENHAB_CONF/automation/js* folder.
+
+To import and use a local into your rule you need to require it and create a `DateTimeFormatter` that uses it.
```
florian-h05, line 843:
```suggestion
To import and use a local into your rule you need to require it and create a `DateTimeFormatter` that uses it:
```

README.md
```diff
@@ -829,6 +829,24 @@ actions.Exec.executeCommandLine(time.Duration.ofSeconds(20), 'echo', 'Hello Worl
...
+
+```javascript
+var Locale = require('@js-joda/locale_de-de').Locale.GERMAN;
+var formatter = time.DateTimeFormatter.ofPattern('MM/dd/yy HH:mm').withLocale(Locale);
```
florian-h05, line 847:
```suggestion
var formatter = time.DateTimeFormatter.ofPattern('dd.MM.yyyy HH:mm').withLocale(Locale);
```

This should reflect something like 05.02.2024, which is the common notation for dates in Germany, correct?

florian-h05 (changes requested):
In general LGTM, thanks.
Just a few wording and styling comments,

README.md
rkoshak: I have no idea what the common notation is. If that's what it is in Germany we should go with that. I think I intended to come back and change it because the current format shown is the messed up US formatting.

README.md
rkoshak: I accepted the rewording but cut off the "of it" part. I don't think it's needed.

florian-h05 (approved):

florian-h05:
@rkoshak Thanks, looks good now! This finally fixes #54.

PR #331, opened 2024-05-01
[items] ItemHistory: Adjust to breaking changes & Add new methods for future states

florian-h05:
Refs https://github.com/openhab/openhab-core/pull/3736.

- Rename `Item::history` to `Item::persistence`.
- Rename `HistoricItem` to `PersistedItem` and introduce `PersistedState` for state as string, Quantity & number.
- Make variance, deviation, average, sum & delta return the state as `PersistedState` instead of number.
- Add `**Until` methods for all existing queries.
- Add `nextState` and `nextUpdate` methods.
- Rename `historicState` method to `persistedState`.
- Remove `latestState` method.
- Remove deprecated `evolutionRate` method.
- Add `removeAllStatesSince`, `removeAllStatesUntil` & `removeAllStatesBetween` methods.

With this PR, support for accessing persistence on openHAB 4.1.x or older will be dropped with the next openhab-js release.
Users should either upgrade openHAB to 4.2.x or stay with the current 4.x.x series of openhab-js.
Since there are no known issues in openhab-js, staying on 4.x.x should be fine.


florian-h05:
TBD after merge: 

- [ ] Update Blockly code generation.
- [ ] Update openhab-js tern defs.

rkoshak:
> Make variance, deviation, average, sum & delta return the state (including unit) as string instead of a number.

For consistency, does it make sense to offer `numberAverage` and `quantityAverage` (et. al.) same as we have in Item for accessing the state? `HistoricState` already has those options too so it feels reasonable to expect there to be options for these persistence queries too.

jlaur:
> * Make variance, deviation, average, sum & delta return the state (including unit) as string instead of a number.

Would it make sense to return a `Quantity` instead of a number? I guess in most cases the returned string would be used for creating a `Quantity`?

jlaur:
I wanted to ask this question for some hours, but didn't find time until now, and just after posting I saw @rkoshak's similar question. 😄

rkoshak:
> Would it make sense to return a Quantity instead of a number?

I don't think we'd want to always return a Quantity. If the Item is a plain Number Item there is no unit so a Quantity wouldn't be appropriate in that case. 

Returning a String by default with options for numeric and quantity maps to other places in the library where a state of an Item is accessed (from the Item or from a HisotricState).

florian-h05:
My proposal is to create a class `PersistedState` or `HistoricState` inside the helper library, which has the same properties as https://openhab.github.io/openhab-js/items.HistoricItem.html, and have the current `HistoricItem` just extend this class.
This way it is easily possible to get the state in different formats.

WDYT?

rkoshak:
Works for me.

florian-h05:
Implemented and tested :+1:

PR #332, opened 2024-05-02
Remove deprecated methods & properties

florian-h05:
- [rules] Remove deprecated `state` and `receivedTrigger` event object properties.
- [cache] Remove deprecated default cache (users should either use private or shared cache).
- [actions] Remove `createTimerWithArgument` (users should use `createTimer` or `setTimeout` instead, `createTimerWithArgument` is not even thread-safe).

All these have been marked as deprecated at least since openHAB 3.4, which was released end of 2022, which means users have had enough time to adjust their code.

@rkoshak FYI. I have checked openhab-rules-tools, only usage of deprecated methods I could find was in open_reminder.yaml and debounce.yaml. However both of these are only published for openHAB 3.x on the marketplace, therefore this should be fine.


rkoshak:
I want to make sure that those versions of the rule templates remain compatible with the version of the helper library that comes with the 3.4 add-on so I'll leave them as is. 

I agree, plenty of time has passed. There shouldn't be any OH 4 users trying to use these.

florian-h05:
> so I'll leave them as is.

Sounds reasonable :+1: 

> I agree, plenty of time has passed.

And moving away from the deprecated methods was relatively easy and straighforward.

I am getting rid of the deprecated stuff now because with #331 there already is a breaking change for the next release, so I can now bundle all breaking changes into one release.

Issue #333, opened 2024-05-30
Support passed in values for manually called JSRules


rkoshak:
I've had a few users who want to use my rule templates but who also want to use .js files and JSRule for all their other rules. This means that they have a rule that gets called from a UI rule and when called data is passed to the called rule. 

I UI rules one can simply use the passed in variables by name. Clearly that won't work with JSRule.

I've looked at how the `event` Object is created: https://github.com/openhab/openhab-js/blob/523d874a192d97444f502c271e3ed32b8bc45c63/src/rules/rules.js#L358 and it's not clear to me what the last line of that function is doing. What's `"module"` and does that contain the additional passed in data?

If not, would it be acceptable to do one of the following?

1. if there is no rule trigger (i.e. it was manually triggered), stuff `data` with anything extra that's in `input`
2. no matter how the rule is triggered, add anything extra in `input` to `data`
3. add `input` itself to `data` as `raw`

With option 1 and 2, if the calling rule passes a variable `foo` the callee rule would access that via `event.foo`.

With option 3, if the calling rule passed a variable `foo` the callee rule would access that via `event.raw.foo` (or something more appropriate if `raw` is the wrong name). 

Or maybe `module` already does option 3? I assume this is the result from https://www.openhab.org/javadoc/latest/org/openhab/core/automation/rule#getModules() and the docs for `Module` talk about `Inputs` but as far as I can tell this is not what I'm looking for. 

Before I go too far down the rabbit hole researching and prepare a PR I want to make sure it would be accepted in the first place. I'm inclined to use approach 3 as the least likely to cause problems now and in the long run. 

## Your Environment
openhab-js 5.0.0 is my current environment.

florian-h05:
Let's go for solution 3!
Just don't forget to update the README.

PR #334, opened 2024-05-31
[rules] Add original input to event for file-based rules

rkoshak:
# [JSRule] Add original input from the event that triggered a rule to the JS `event` Object passed to file based rules

# Description

There are circumstances where one may need to access the raw Java Objects from the event passed to a rule or to access variables passed into the rule by a calling rule.  This PR adds the original contents of all the input passed to the rule as a `raw` property on the `event` Object passed into file based rules. Managed script actions and script condition always get the original and not the converted JS `event` Object. 

For a usage example, to access the property `foo` passed from a calling rule, use `event.raw.foo` or `event.raw['foo']`.

# Testing

To test I created a JS Rule that dumps the `event` Object.

```javascript
rules.JSRule({
  name: 'Dump event Object',
  description: 'Logs out the event Object',
  triggers: [],
  execute: (event) => {
    utils.dumpObject(event, true);
  },
  id: 'dumpEvent',
  tags: ['Script']
});
```

I then tested by manually running the following from the -Scratchpad- Script:

```
rules.runRule('testEvent', {'foo':'bar', 'baz':'bim'}, true);
```

and verified that `event.raw.foo` is `"bar"` and `event.raw.baz` is `"bim"`.

Next I exercised it from a rule that uses Thing trigger by instantiating an instance of https://community.openhab.org/t/thing-status-reporting-4-0-0-0-4-9-9-9/143180/ that calls this rule when ever a Thing changes status with a number of properties passed in. I then verified that the `event.raw` contains `thingID`, `oldStatus`, `oldDetail`, `newStatus`, `newDetail` and `thing` and each is the expected value. This second test was to ensure that the data gets passed both when a rule is called from a manually triggered rule and from an event triggered rule. 

Solves #333


florian-h05 (approved):
LGTM, thanks!

I will update the JSDoc and regenerate type defs and merge afterwards.-

PR #371, opened 2024-08-06
[items] Item: Add increaseValueTo/decreaseValueTo methods

### LINKED ISSUE #369: Function to send Item updates only in some cases
I have a rule "when it gets dark outside, turn on the lights to 20%": `items.getItem("Lights").sendCommand("20")`. However this will darken the light when it has been turned to 100% manually before. I wrapped it in a `if` to avoid this, but maybe a function for this could be added, similar to `sendCommandIfDifferent()`:
- `increaseValueTo(value)`: Send the command only when `value` is larger than the current state.
- `decreaseValueTo(value)`: Send the command only when `value` is smaller than the current state.

This only applies to `Number` Items.

Feel free to close this if you think it's too niche.

florian-h05:
Closes #369.

Also:
- Make `Item::getToggleState` method private.
- Adjust `Item::getToggleState`/`Item::sendToggleCommand` method to .type changes from a previous PR
- Use brightness value of Color Items for `Item::numericState`.


florian-h05:
@mueller-ma You can download the dist-without-license artifact from https://github.com/openhab/openhab-js/actions/runs/10271819868?pr=371 and put the openhab.js file contained in the ZIP into the $OPENHAB_CONF/automation/js/node_modules folder.
Then disable injection caching in the add-on settings and you can try out the new methods.

WRT to naming: increaseValueTo and decreaseValueTo are short, but not really in line with sendCommand, sendToggleCommand, sendCommandIfDifferent. But sendIncreaseValueToCommand are too long?

mueller-ma:
What about sendIncreaseCommand()?

florian-h05:
Sounds good - have you been able to give this PR a try?

mueller-ma:
I tested `sendIncreaseCommand(10)` on a Dimmer item that was at 100 and it got decreased to 10. How can I help to debug this?

florian-h05:
I cannot reproduce this issue with basic Dimmer and Number Items (means no metadata, just a Item with a name).

If you have changed the dimmer to 100 and sendIncreaseCommand right afterwards, i.e. by having a script that does these two steps after each other, it may be the case that a wrong command is sent because the Item state hasn't been updated yet.

This works just fine for me:

```js
var item = items.test_dimmer

item.sendCommand(100)

java.lang.Thread.sleep(500)
item.sendIncreaseCommand(10)

java.lang.Thread.sleep(500)
item.sendDecreaseCommand(15)

java.lang.Thread.sleep(500)
item.sendIncreaseCommand(50)

java.lang.Thread.sleep(500)
item.sendIncreaseCommand(10)
```

Note the sleeps to avoid the above mentioned issue.

florian-h05:
You can enable debug logging (just added it) by setting the log level of `org.openhab.automation.openhab-js.items` to DEBUG.

florian-h05:
@mueller-ma Let me know when you have been able to test it again.

mueller-ma:
Now it seems to work. Maybe I did something wrong with downloading the library the first time.

Thank you for this PR 👍

Issue #375, opened 2024-08-26
[items] ItemPersistence: Add median functions 


florian-h05:
See https://github.com/openhab/openhab-core/pull/4345.

PR #376, opened 2024-08-26
[items] ItemPersistence: Add medianSince, medianUntil and medianBetween methods

mherwege:
Closes https://github.com/openhab/openhab-js/issues/375

Signed-off-by: Mark Herwege <mark.herwege@telenet.be>


mherwege:
Not sure why the signoff check fails

src/items/item-persistence.js
```diff
@@ -602,6 +602,45 @@ class ItemPersistence {
     return _stateOrNull(PersistenceExtensions.averageBetween(this.rawItem, ...arguments));
   }
 
+  /**
+   * Gets the median value of the state of a given Item since a certain point in time.
+   *
+   * @example
+   * var yesterday = new Date(new Date().getTime() - (24 * 60 * 60 * 1000));
```
florian-h05, line 609: Rather use JS-Joda in the example - we never use the native JS Date because it is not that great:

```suggestion
   * var yesterday = time.toZDT().minusDays(1);
```

florian-h05 (changes requested):
Just one minor remark:

(Please regenerate type defs after that change.)

src/items/item-persistence.js
mherwege: Changed

florian-h05 (approved):
LGTM, thanks!

PR #397, opened 2024-11-06
Update return type for getItem() to indicate it can return null

Dopeyr:
# Update items.getItem return type to indicate it can return null

# Description

getItem() can return null when `nullIfMissing` is set to true, but the return type is currently just `Item`.

# Testing

N/A


florian-h05:
This is intentional, as by default null is not returned and instead an error is thrown.
Declaring null as output leads to TypeScript type checking inside the IDE to require checking for null when calling `getItem(itemName)`, which is not intended.

Dopeyr:
> This is intentional, as by default null is not returned and instead an error is thrown. Declaring null as output leads to TypeScript type checking inside the IDE to require checking for null when calling `getItem(itemName)`, which is not intended.

Thanks @florian-h05 - I did wonder what the solution to this is as `getItem(itemName, true)` returns me the wrong type and I'm working round it with `getItem(itemName, true) as Item | null` in typescript.

I looked into `@overload` in jsdoc to specify two function signatures (if nullIfMissing is not specified, return `Item`, if it is specified return `Item | null`) but I wasn't successful. The types output didn't change. I'm new to jsdoc so I could just be missing something simple there.

Thanks, Andy.

florian-h05:
Checking https://www.typescriptlang.org/docs/handbook/jsdoc-supported-types.html, it doesn’t seem there is a solution for that, which is unfortunate but can be lived with.

PR #408, opened 2024-12-18
Use configured time-zone when converting `DateTimeType` to `ZonedDateTime`

### LINKED ISSUE #404: Adjust use of deprecated `DateTimeType::getZonedDateTime`  method in `time::toZDT`
Due to https://github.com/openhab/openhab-core/pull/3583,
it is required to adjust 

https://github.com/openhab/openhab-js/blob/5ebf646404b75de3c26b2d810f25520e3843ca12/src/time.js#L189

and 

https://github.com/openhab/openhab-js/blob/5ebf646404b75de3c26b2d810f25520e3843ca12/src/time.js#L294

to `DateTimeType.getZonedDateTime(zoneId)`.

This however cannot be done without breaking backwards compatibility with openHAB < 4.3, so we have to postpone this change to the next breaking openhab-js release.

jlaur:
This removes usages of the deprecated `DateTimeType` method [getZonedDateTime()](https://www.openhab.org/javadoc/latest/org/openhab/core/library/types/datetimetype#getZonedDateTime()) and instead applies the time-zone configured in openHAB.

Related to openhab/openhab-core#3583
Resolves #404

### DIFF
--- a/src/time.js
+++ b/src/time.js
@@ -186,7 +186,7 @@ function _convertItemRawState (rawState) {
   } else if (rawState instanceof StringType) { // String type Items
     return _parseString(rawState.toString());
   } else if (rawState instanceof DateTimeType) { // DateTime Items
-    return javaZDTToJsZDT(rawState.getZonedDateTime());
+    return javaInstantToJsInstant(rawState.getInstant()).atZone(time.ZoneId.systemDefault());
   } else if (rawState instanceof QuantityType) { // Number:Time type Items
     return _addQuantityType(rawState);
   } else {
@@ -288,10 +288,10 @@ function toZDT (when) {
     return _addMillisToNow(when.floatValue());
   }
 
-  // DateTimeType, extract the javaZDT and convert to time.ZDT
+  // DateTimeType, extract the javaInstant and convert to time.ZDT, use the configured timezone
   if (when instanceof DateTimeType) {
-    log.debug('toZTD: Converting DateTimeType ' + when);
-    return javaZDTToJsZDT(when.getZonedDateTime());
+    log.debug('toZDT: Converting DateTimeType ' + when);
+    return javaInstantToJsInstant(when.getInstant()).atZone(time.ZoneId.systemDefault());
   }
 
   // Add Quantity or QuantityType<Time> to now


src/time.js
```diff
@@ -186,7 +186,7 @@ function _convertItemRawState (rawState) {
   } else if (rawState instanceof StringType) { // String type Items
     return _parseString(rawState.toString());
   } else if (rawState instanceof DateTimeType) { // DateTime Items
-    return javaZDTToJsZDT(rawState.getZonedDateTime());
+    return javaInstantToJsInstant(rawState.getInstant()).atZone(time.ZoneId.systemDefault());
```
florian-h05, line 189: I would rather get the ZonedDateTime from the DateTimeType with Java code and convert that Java ZDT to a JS ZDT.

jlaur: You cannot get it from the `DateTimeType` without breaking compatibility with OH 4.3 - see https://github.com/openhab/openhab-js/issues/404#issuecomment-2536814538. But you can of course get it from the Java `Instant` class in the same way as `DateTimeType` does it:

```java
    public ZonedDateTime getZonedDateTime(ZoneId zoneId) {
        return instant.atZone(zoneId);
    }
```

So in this case we would need to:
- Convert `time.ZoneId.systemDefault()` to Java `ZoneId`.
- Call Java `rawState.getInstant.atZone`.
- Convert the result to JS `ZonedDateTime`.

What would be the benefit of doing that rather than doing it directly in JS?

florian-h05: Possibly performance impact, but I guess it doesn’t matter really matter given how the javaZDTToJsZDT method works.

florian-h05 (approved):
LGTM, thanks!

Have you tested this?

jlaur:
> Have you tested this?

Yes, given this:
```shell
openhab:update TestDateTime 2024-10-31T12:00:01-0800
```

and this rule code:
```javascript
var zoned = time.toZDT(items.TestDateTime);
console.log(zoned.toString());
```

Initial:
```
2024-12-23 21:00:00.186 [INFO ] [enhab.automation.script.file.test.js] - 2024-10-31T21:00:01+01:00[Europe/Copenhagen]
```

After setting OH time-zone and running the rule again:
```
2024-12-23 21:04:04.607 [INFO ] [enhab.automation.script.file.test.js] - 2024-10-31T18:00:01-02:00[America/Scoresbysund]
```

florian-h05:
Thanks for taking care of this!

Issue #392, opened 2024-10-28
types/triggers.d.ts should not expose its own Item definition


wertzui:
After upgrading to the latest version, VS code tells me that there is a problem when I pass an `Item` to `triggers.ItemStateChangeTrigger`, because `Item` does not expose `"__#5@#getToggleState"(): "PAUSE" | "PLAY" | "OPEN" | "CLOSED" | "ON" | "OFF"` which is defined here
https://github.com/openhab/openhab-js/blob/7f2b7a41a5d68e0464a05dc9654799ffedd35abf/types/triggers.d.ts#L36 and got introduced with https://github.com/openhab/openhab-js/commit/2348409e0a6af49c133b0438578b4f8c6423b3d1.

Besides this method definition being obviously wrong (It should probably be `getToggleState(): "PAUSE" | "PLAY" | "OPEN" | "CLOSED" | "ON" | "OFF"`), I think `triggers.d.ts` should not expose its own `Item` definition, but rather use the one from `items.d.ts`.

florian-h05:
#394 fixes the issue with `"__#5@#getToggleState"(): "PAUSE" | "PLAY" | "OPEN" | "CLOSED" | "ON" | "OFF"` by not making this method private anymore.

> I think triggers.d.ts should not expose its own Item definition, but rather use the one from items.d.ts.

Yes, but our type definitions are emitted by the TypeScript compiler, which currently inlines type definitions.
This is an upstream issue: https://github.com/microsoft/TypeScript/issues/37151

Dopeyr:
This might be solvable with Typescript 5.5, ref: https://github.com/microsoft/TypeScript/issues/22160

Adding

```/**
 * @import { Item } from './items/items'
 */```

to `triggers.js` no longer adds the export of the `Item` type, so vscode does not suggest `Item` from triggers.js. 

However, the resulting type def in in `types/triggers.d.ts` is:

```/**
 * Creates a trigger that fires upon an Item changing state.
 *
 * @example
 * ItemStateChangeTrigger('my_item'); // changed
 * ItemStateChangeTrigger('my_item', 'OFF', 'ON'); // changed from OFF to ON
 * ItemStateChangeTrigger('my_item', undefined, 'ON'); // changed to ON
 * ItemStateChangeTrigger('my_item', 'OFF', undefined); // changed from OFF
 *
 * @memberof triggers
 * @param {Item|string} itemOrName the {@link Item} or the name of the Item to monitor for change
[...]

PR #433, opened 2025-04-23
[items] Item: Add numeric and quantity variants for previousState

mherwege:
I was looking into adding this in Blockly. It would be very useful for Blockly to have `previousNumericState` and `previousQuantityState` as well. That way this can be used easily to replace calling persistence methods.


mherwege:
I am struggling with typescript. When I run `tsc --project ./build/tsconfig.json` there is a funny change introduced in `quantity.d.ts`. The online build fails on that. I can remove that funny change manually, removing the online build error on that file, but the of course the `quantity.d.ts.map` file is not correct in that case. I have no idea why this happens in the first place.

florian-h05:
I think this is #392, being a bug in the typescript compiler itself (see #392 for details).

mherwege:
I just pushed it with all type generation done by the command. Look at `quantity.d.ts`. Suddenly it contains extra things onlines 7 to 10:
```
    readonly label: string; /**
     * @type {QuantityType}
     * @private
     */
```
I have no clue where this comes from. The online build then complains the `quantity.d.ts` and  `quantity.d.ts.map` files have changed.
So while it may be related to the issue you linked, I am not sure it is. And I have no clue how to solve that.

src/items/items.js
```diff
@@ -170,8 +170,45 @@ class Item {
...
+    return this.previousRawState == null ? null : this.previousRawState.toString();
+  }
+
+  /**
+   * Numeric representation of Item previous state, or `null` if state is not numeric or not available.
+   * @type {number|null}
+   */
+  get previousNumericState () {
+    const previousState = this.previousRawState == null ? null : this.previousRawState.toString();
+    if (previousState == null) return null;
```
florian-h05, line 182: Why doing an additional null check here? 
Can toString return null?

Please also have a look at special handling of Color Items, see numericState above.

src/items/items.js
```diff
@@ -170,8 +170,45 @@ class Item {
...
+    return isNaN(numericState) ? null : numericState;
+  }
+
+  /**
+   * Previous item state as {@link Quantity} or `null` if state is not Quantity-compatible, Quantity would be unit-less (without unit) or not available.
+   * @type {Quantity|null}
+   */
+  get previousQuantityState () {
+    const previousState = this.previousRawState == null ? null : this.previousRawState.toString();
+    if (previousState == null) return null;
```
florian-h05, line 193: Why doing an additional null check here?

florian-h05 (commented):
Can you please refactor the code for numeric and quantity state into a private function (declare a function outside of the class with a name starting with `_`) that accepts a raw Java state? This would avoid the code duplication introduced here.

florian-h05:
> I have no clue where this comes from. The online build then complains the quantity.d.ts and quantity.d.ts.map files have changed.

Checking out your branch, my TypeScript reverts this change.

src/items/items.js
mherwege: previousRawState can be null, but I will refactor that.

src/items/items.js
mherwege: previousRawState can be null

mherwege:
> > I have no clue where this comes from. The online build then complains the quantity.d.ts and quantity.d.ts.map files have changed.
> 
> Checking out your branch, my TypeScript reverts this change.

Even after my refactoring, and re-generating the typescript maps, I still see it. Again, I don't know what is going on.

src/items/items.js
florian-h05: Correct, I would simplify the code by first checking for null and early returning instead of checking twice for null.
The toString method shouldn’t return null for a non-null type.

mherwege:
What is this check for broken links that now also fails? It fails on links to commits and I fail to see the relation to this PR.

florian-h05:
The broken link check checks whether links in Markdown files are “alive” or dead. It fails due to GitHub’s rate limit. We can ignore this for now and I will adjust its configuration.

florian-h05 (approved):
LGTM now, thanks!
I have regenerated the types locally, now they look good. Something seems to be wrong with your TS installation.

PR #450, opened 2025-07-22
Update README.md

lsiepel:
Follow-up to: https://github.com/openhab/openhab-addons/pull/18947/files

### DIFF
--- a/README.md
+++ b/README.md
@@ -141,7 +141,7 @@ See [openhab-js](https://openhab.github.io/openhab-js) for a complete list of fu
 Note that `event` object is only available when:
 
   1. the UI based rule was triggered by an event and is not manually run! Trying to access `event` on a manual run does not work (and will lead to an error), use `this.event` instead (will be `undefined` in case of manual run).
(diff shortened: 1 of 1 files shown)


florian-h05 (commented):
Hm, running a simple script with `console.log(event)` manually gives back `Execution triggered by manual`, so I think the docs are a bit outdated on this part.
WDYT?

NbodySpecial:
The section is about UI scripts, not console scripts.

florian-h05:
When manually running or calling from another rule, the following code snippet from a UI-based script (e.g. scratchpad), I get the following logged:

```js
console.log(event)
console.log(event.itemName)
console.log(event.type)
```

```
11:46:57.106 [INFO ] [enhab.automation.script.ui.scratchpad] - Execution triggered by manual
11:46:57.107 [INFO ] [enhab.automation.script.ui.scratchpad] - undefined
11:46:57.108 [INFO ] [enhab.automation.script.ui.scratchpad] - ExecutionEvent
```

So `event` is obviously available when running manually, the suggested docs change is hence wrong.
We can agree that we need to update the UI event object properties table, since `itemName` is of cource not available for all trigger types, but event object is still available.

PR #479, opened 2025-10-02
README: clarification for many rules from the same .js file

dilyanpalauzov:
Same as and as discussed in https://github.com/openhab/openhab-addons/pull/19410.


florian-h05:
This needs to be adjusted once I merge #490.
As I anyway need to write docs for it, I can pick up the suggested change from this PR then.
Please do not close this PR yet.

PR #487, opened 2025-10-07
README: table Raw UI Event Object formatting

dilyanpalauzov:


florian-h05:
Fill fix this when working on the docs locally, thanks!

PR #490, opened 2025-10-08
[rules] JSRule: Add support for creating "standard" rules (not SimpleRule)

florian-h05:


florian-h05:
@Nadahar This enables JSRule to create "standard" openHAB rules in addition to its ability of creating SimpleRules.
Please let me know what you think about the naming of the option that enables this `dedicatedContext` and the JSDoc documentation for it.

florian-h05:
FYI you may use this for testing:

```js
const { rules, triggers, utils } = require('openhab')

function myFunction() {
  console.log('myFunction()')
}

rules.JSRule({
  name: 'Shared Context Rule',
  execute: (event) => {
    myFunction()
    utils.dumpObject(this)
  }
})

rules.JSRule({
  name: 'Dedicated Context Rule',
  triggers: triggers.ItemCommandTrigger('test_switch'),
  execute: (event) => {
    myFunction() // expect this to fail
    utils.dumpObject(this)
    java.lang.Thread.sleep(5000)
  },
  dedicatedContext: true
})
```

Nadahar:
> @Nadahar This enables JSRule to create "standard" openHAB rules in addition to its ability of creating SimpleRules.
> Please let me know what you think about the naming of the option that enables this `dedicatedContext` and the JSDoc documentation for it.

Please give me some time to study it, I'm "slow" when it comes to these things 😉

Nadahar:
I added a comment to one of the CoPilot comments that have been "resolved". I'm not sure if you'll see that or not, so this is a heads-up.

florian-h05:
Just let me know once you are finished with testing.

Nadahar:
About the naming, I think it looks good 👍

Nadahar:
Did you ever find my comment about how to attach the source?

> Actually, there is a better way to do this now. I've just never gotten around to dealing with it. When I added the "Source" tab for rules in MainUI, I also made a new way to attach the source with the rules. Instead of attaching it to the first Action, it can be attached to the Rule itself. But, if it can't find anything there, it "falls back" to checking the first Action, so that nothing would break. Still, it would be better/more correct to attach the source to the rule itself, which would also eliminate this potential corner case.

https://github.com/openhab/openhab-webui/blob/892ad5006688b12f1adfaaf5a71c067bf9821763/bundles/org.openhab.ui/web/src/pages/settings/rules/rule-edit.vue#L963-L976

I always made to raise this anyway, so when I saw the code here, I might as well mention it now. It's very little change, basically, you move it to the Rule instead of the Action, and use the key `source` instead of `script`:
```javascript
  ruleConfiguration.put('type', SCRIPT_TYPE);
  ruleConfiguration.put('source', '// Code to run when the rule fires:\n// Note that Rule Builder is currently not supported!\n\n' + ruleConfig.execute.toString());
```

Nadahar:
This is amazing 🥳 

This opens up the possibility for JS rules that can run in parallel  (if you don't need shared functions) and that can be duplicated in MainUI (if that has any value). But, you could also implement support for making rule templates from JS (if there is any point in doing that, of course) using this technique.

The approach could also be used to enable scripted conditions, also for `SimpleRule`s (but the `Condition` would still be in a dedicated context). It can probably also be used to enabled scripted triggers, if they have any actual use (it's hard to see exactly what they would achieve, given that they'd need some data to "trigger on").

Nadahar:
If you want to take this "far out there", maybe you could find a way to define another "dedicated context" in the same file where shared functions could be defined. This "dedicated context" could then be "injected" into the other dedicated context, for example by registering it as a `Source` in the GraalJSEngine. That way, both Actions and Conditions could utilize the shared functions (and in theory, other JS source files as well).

Just disregard if you're currently not looking for a challenge, this is just ideas that pop into my head of what *might be possible* to achieve 👼

src/rules/rules.js
```diff
@@ -203,45 +221,86 @@ function setEnabled (uid, isEnabled) {
...
+ * <p>Through the {@link RuleConfig} it can be opted into creating these rules instead of SimpleRules by the {@link rules.JSRule} function.
+ *
+ * @param {string} ruleUID
+ * @param {RuleConfig} ruleConfig
+ * @return {HostRule} the created {@link https://github.com/openhab/openhab-core/blob/main/bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/internal/RuleImpl.java RuleImpl}
+ * @private
+ */
+function _createRule (ruleUID, ruleConfig) {
+  let script = ruleConfig.execute.toString();
+  script = script.match(/(function)?[^{]+\{([\s\S]*)}$/);
```
Nadahar, line 243: So, this is where "the magic" happens? I don't quite get that you can extract the source code just by calling `toString()` on the function, but it's great that you can.

PR #491, opened 2025-10-09
README and rules.js: reword documentation

dilyanpalauzov:
Make consistent no-hyphen policy for the terms “UI based” and “File based” rules and scripts.
    
Spell explicitly that parallel execution is blocked only when many rules originate from the same .js file.
    
Use on more places “separate dedicated” together for context, otherwise the question arises what is a separate context, which is not a dedicated context.  I do not know if there should be a comma between separate and dedicated, but whether there is a comma should be consistent.

### DIFF
--- a/README.md
+++ b/README.md
@@ -178,7 +178,7 @@ In case the event object does not provide type-conversed properties for your cho
 
 See [openhab-js : EventObject](https://openhab.github.io/openhab-js/global.html#EventObject) for full API documentation.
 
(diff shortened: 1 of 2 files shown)


florian-h05 (commented):
Why `UI based` and not `UI-based`?
I am not sure what is the better term …

dilyanpalauzov:
I have no preference for UI based or UI-based, you can consult others.  Currently it is sometimes written so and sometimes the other way.

PR #504, opened 2025-12-26
[items] Item: Add `isInitialized` property

### LINKED ISSUE #503: Proposal to introduce `Item`.`isInitialized`
This is somewhat counter-intuitive because of double negation:

```javascript
if (!items.WashingMachine_WaterConsumption.isUninitialized) {
```

Why:
- Violates expectation: `isX` usually reads as a positive state.
- Harder to scan: You have to mentally flip the meaning.
- Leads to awkward logic: `!isUninitialized` is semantically clumsy.

I would therefore propose to introduce `isInitialized` or `initialized`.

Ideally `isUninitialized` should be deprecated, at least eventually, but probably it can never be removed because it would break scripts.

## Your Environment

- openHAB version used (check Main UI -> About): 5.1
- openhab-js version used (run `console.log('openhab-js ' + utils.OPENHAB_JS_VERSION)` from a JS script): 5.16.1

jlaur:
Resolves #503


jlaur:
Some changes from `npm run types` are unrelated and should probably be extracted to a separate PR in case they belong to merged bugfixes that should be backported. I'm not sure why the build failed.

jlaur:
> The generated types sometimes look very weird, but are all caused by the changes in your PR.
> During the build, type generation is run as and if it changes a file, the build fails. You have to commit all changes.

Build failed with:
```
Error: Files have changed: types/items/itemchannellink.d.ts types/items/itemchannellink.d.ts.map types/rules/trigger-builder.d.ts types/rules/trigger-builder.d.ts.map
```

These files are all included already. I need to check my node.js version, perhaps it's too old.

jlaur:
> I need to check my node.js version, perhaps it's too old.

Actually it was too new. I tried to switch from 23.4.0 to 22.17.1 as mentioned in `.nvmrc` and run `npm run types` again, but it didn't change anything.

jlaur:
> Build failed with:
> 
> ```
> Error: Files have changed: types/items/itemchannellink.d.ts types/items/itemchannellink.d.ts.map types/rules/trigger-builder.d.ts types/rules/trigger-builder.d.ts.map
> ```
> 
> These files are all included already.

I now tried to exclude these files from this PR to see if the problem is the other way around, but it still fails. In #505 I updated these files from **main**, but that build also fails.

PR #527, opened 2026-05-25
[rules] Collapse input map in _getTriggeredData() 

Nadahar:
https://github.com/openhab/openhab-core/pull/4919 has caused loss of data in the `ctx` object that is sent to action handlers when executing actions, because by removing the prefix, keys with the same name from multiple modules will overwrite each other, and you don't actually know what objects you end up with (because maps are unordered, this can happen in any order).

The assumption that the prefix isn't needed is only true when there is only one previous module "in the chain", which is the most common situation, but still isn't the *only* possible situation. Examples of situations where this might not be true are multi-action rules or rules that are started by other rules (the context of the "triggering rule" is also in the `ctx` map).

I've tried to do the "collapsing" of the map in `_getTriggeredData()` so that https://github.com/openhab/openhab-core/pull/4919 can be reverted and the information restored.

Whether the collapsed or the "full" map should be stored in the `raw` property is a judgment call, for simple rules, it's "easier" for the user to not have to take the module ID into account, while it will make it impossible to handle more complex scenarios. I would recommend keeping the "full" map in `raw` and then use `EventObject` for the simplified case.

Since GraalJS won't run under Equinox (I suspect that the reason is that dependencies depend on Karaf), I can't test this. So, I can give no guarantees, but it should at least be easy to tweak this into something that does work.


Nadahar:
@florian-h05 I've modified this PR so that now the "raw" map should contain both the prefixed and the unprefixed entries. I'm not sure how to replace the helper lib with "my version" for local testing though.

Nadahar:
I can't understand that it should make a difference, but I also made the extraction of `command`, `oldState` etc. use the collapsed map for extraction. As far as I can tell, with this code, there should be zero possibility for any kind of regression for JS users if core is changed back.

florian-h05:
I am reviewing your PR right now, will push a few minor changes (mainly stylistic) soon.

florian-h05 (approved):
LGTM, thanks!

PR #529, opened 2026-06-01
[rules] _getTriggeredData: Provide a merge function when collapsing input map

Nadahar:
This should fix the merge conflict introduced by #527.


Nadahar:
@florian-h05 I guess this is a bit of a rush, if clients automatically upgrade.

florian-h05:
Clients never upgrade automatically. What is shipped with the add-on has to be defined in the POM.
What users do with their npm is their thing and out of our control. openHABian might offer to upgrade, but it is asking, not doing IIRC.

florian-h05 (approved):
Many thanks for the quick fix.
I can merge this PR from my phone, for the release I need my laptop which might take a few hours.

PR #536, opened 2026-07-09
[utils] Introduce javaify() function

Nadahar:
This is more of an idea than a finished PR. The code is untested, I don't know if it will ever work. The point is to suggest an approach for how to "convert" JS objects to Java objects.

Something similar could be made to convert things the other way, but that would have to be called by the user on demand only, because in some situations, it's probably preferable to retain the Java objects.

If the code works as intended, it will recursively convert Maps, Sets, Arrays and objects of any depth. As far as I understand it, there's no need for a special handing for strings because they will automatically be converted to Java strings by Graal - but I'm not entirely sure if that's true. If it isn't true, the function would need to create Java Strings as well.


src/rules/rules.js
```diff
@@ -194,7 +192,93 @@ function runRule (uid, args = {}, conditions = true) {
...
+function javaify (val) {
+  if (val === null || val === undefined) {
+    return null;
+  }
+
+  if (Java.isHostObject(val)) {
+    return val;
+  }
+
+  // Convert js-joda objects
```
rkoshak: I think js-joda Objects are already converted to Java in the add-on. I could be wrong about that though,

src/rules/rules.js
```diff
@@ -194,7 +192,93 @@ function runRule (uid, args = {}, conditions = true) {
...
+    }
+    if (typeName === 'LocalDateTime') {
+      return Java.type('java.time.LocalDateTime').parse(val.toString());
+    }
+    if (typeName === 'Instant') {
+      return Java.type('java.time.Instant').parse(val.toString());
+    }
+  }
+
+  if (typeof val === 'function') {
```
rkoshak: I say reject with a warning.

src/rules/rules.js
```diff
@@ -194,7 +192,93 @@ function runRule (uid, args = {}, conditions = true) {
...
+  if (typeof val !== 'object') {
+    return val;
+  }
+
+  // Convert JavaScript Date
+  if (val instanceof Date) {
+    return Java.type('java.time.Instant').ofEpochMilli(val.getTime());
+  }
+
+  // Convert arrays
```
rkoshak: There is already a function that does this in utils.

src/rules/rules.js
```diff
@@ -194,7 +192,93 @@ function runRule (uid, args = {}, conditions = true) {
...
+  if (Array.isArray(val)) {
+    const ArrayList = Java.type('java.util.ArrayList');
+    const list = new ArrayList();
+    for (const element of val) {
+      list.add(javaify(element));
+    }
+    return list;
+  }
+
+  // Convert JS Maps
```
rkoshak: There is already a function that does this in utils.

src/rules/rules.js
```diff
@@ -194,7 +192,93 @@ function runRule (uid, args = {}, conditions = true) {
...
+  if (val instanceof Map) {
+    const LinkedHashMap = Java.type('java.util.LinkedHashMap');
+    const javaMap = new LinkedHashMap();
+    val.forEach((value, key) => {
+      javaMap.put(key, javaify(value));
+    });
+    return javaMap;
+  }
+
+  // Convert JS Sets
```
rkoshak: There is already a function that does that in utils.

src/rules/rules.js
Nadahar: Yes, I know, but it's easy to do it here, this function is called recursively so I see no need to involve "external code".

Nadahar:
Regarding using existing implementations for various types, I'll leave that up to @florian-h05 to evaluate. I kind of like having it all "under control" here, without being impacted by potential bugs in external code. Also, it's faster to do it inline than to call functions, but the difference might be negligible. The main "reason" for me to do it all here, is to have "control" over the whole conversion in one place, so that it's easy to get an overview over exactly what happens.

Nadahar:
I just remembered, the iterations of collections can't call external functions, the whole point is that this function is recursive - it calls itself also on all the element, so that it does a "deep" (not a shallow)  conversion.

src/rules/rules.js
```diff
@@ -194,7 +192,93 @@ function runRule (uid, args = {}, conditions = true) {
     throw Error('Rule ' + uid + ' is UNINITIALIZED');
   }
 
-  ruleManager.runNow(uid, conditions, new HashMap(args));
+  ruleManager.runNow(uid, conditions, javaify(args));
+}
+
+function javaify (val) {
```
florian-h05: This function should go into utils

src/rules/rules.js
florian-h05: The host type mapping (IIRC it’s named like that) only works if it matches against a specific type in a Java method signature. 
For example, when a parameter takes a Java ZDT, it automatically converts from JS-Joda to Java ZDT.
If the Java method however takes an object, I don’t think a conversion will happen.

Nadahar: It doesn't "cost much" to do it here as well. I just threw in some things I thought might be handy, this function could be evolved to include other "special types" typically used as well.

Nadahar:
In the last commit, I throw an error on `function`. Is that the correct way to reject it?

rkoshak:
> I kind of like having it all "under control" here, without being impacted by potential bugs in external code.

It's not like it's some third party library implemented by others. It's part of openhab-js, written by us, and if you move this function to utils, they are even in the same file.

Nadahar:
> It's not like it's some third party library implemented by others. It's part of openhab-js, written by us, and if you move this function to utils, they are even in the same file.

Yes, I know that, but, unlike many others, I think readability trumps reuse in cases where it's just a little bit of code. As soon as you have to start jumping around the code to see what this and that is doing, it's much harder to get the overview. And, there's the performance benefit.

But, as I said above, it's moot here anyway, because these are recursive. You can't call the other functions, because they don't call `javaify()` on the elements.

Nadahar:
@florian-h05 Isn't this a bit of a hurry, considering that there are plenty of users out there with rules that no longer work? If we can figure this out and make an updated version of openhab-js available, at least we can tell them to upgrade the helper lib manually.

I've done what I can with my limited JS knowledge here, I think the approach should be sound, but there might be details that are off. So, I can't quite take it any further without some input.

When it comes to doing the "reverse conversion", I'm not quite sure how it would look. You already JSify the event object, but we'd need a more generic approach that could JSify "anything", is that right? And, if so, isn't the only sensible way to handle it that users can call that themselves, so that they can also access the Java objects should they need to?

If it's done this way, do we also need an additional Blockly block that can call this function before further processing on JS objects in Blockly?

src/utils.js
```diff
@@ -127,6 +127,88 @@ function javaSetToJsSet (set) {
...
+    if (typeName === 'LocalDateTime') {
+      return Java.type('java.time.LocalDateTime').parse(val.toString());
+    }
+    if (typeName === 'Instant') {
+      return Java.type('java.time.Instant').parse(val.toString());
+    }
+  }
+
+  if (typeof val === 'function') {
+    // What do we do about functions?
```
florian-h05: I'd log a warning and skip them ...

src/utils.js
```diff
@@ -127,6 +127,88 @@ function javaSetToJsSet (set) {
...
+function javaify (val) {
+  if (val === null || val === undefined) {
+    return null;
+  }
+
+  if (Java.isHostObject(val)) {
+    return val;
+  }
+
+  // Convert js-joda objects
```
florian-h05, line 155: To string and string parsing has quite bad performance compared to using epoch milli.

You might want to handle conversion like this:

https://github.com/openhab/openhab-addons/blob/39dee8dbeb1bbd1101ec1312eddf00b31859336b/bundles/org.openhab.automation.jsscripting/src/main/java/org/openhab/automation/jsscripting/internal/OpenhabGraalJSScriptEngine.java#L131-L147

src/utils.js
```diff
@@ -127,6 +127,88 @@ function javaSetToJsSet (set) {
...
+  }
+
+  // Convert JavaScript Date
+  if (val instanceof Date) {
+    return Java.type('java.time.Instant').ofEpochMilli(val.getTime());
+  }
+
+  // Convert arrays
+  if (Array.isArray(val)) {
+    const ArrayList = Java.type('java.util.ArrayList');
```
florian-h05: Already imported at the top.

src/utils.js
```diff
@@ -127,6 +127,88 @@ function javaSetToJsSet (set) {
...
+    const list = new ArrayList();
+    for (const element of val) {
+      list.add(javaify(element));
+    }
+    return list;
+  }
+
+  // Convert JS Maps
+  if (val instanceof Map) {
+    const LinkedHashMap = Java.type('java.util.LinkedHashMap');
```
florian-h05: I'd prefer having these imports at the top and not spread across the file ...

src/utils.js
Nadahar: What would that be? We can't really "skip" parameters here, so that would mean to return `null` as far as I can understand, e.g.

```js
    log.warn('Function parameters aren\'t allowed');
    return null;
```

Users could easily miss this, since function parameters can't work, I think it might be preferable if it just failed, over logging and replacing it with `null`?

florian-h05: Okay - then fail.

src/utils.js
Nadahar: You can't use epoch for `ZonedDateTime` or `LocalDate(Time)`, so that's only viable for one of them - `Instant`. That said, we do string parsing of dates lots of places in OH, I don't expect this to be used on a scale where that really matters. 

Since I don't know how to test the lib, I hope this will work for `Instant`:

```js
    if (typeName === 'Instant') {
      return Java.type('java.time.Instant').ofEpochMilli(val.toEpochMilli());
    }
```

src/utils.js
Nadahar: Is this the correct way "to fail"?

```js
    throw new Error('Functions aren\'t allowed');
```

Nadahar:
@florian-h05 I added another commit where I added `runAsync()` and a return value for `runRule()`. When you wait for a rule to run, you actually get the context in return, including any resulting return values from the execution. It can't hurt to offer to return them? I tried to add docs for this too, but I'm uncertain how to handle the types etc. - so you need to check that extra close.

src/utils.js
Nadahar: Done

src/utils.js
Nadahar: Fixed

src/utils.js
Nadahar: I've changed the `Instant` parsing.

Nadahar:
I guess it's time to take this out of draft.

src/utils.js
florian-h05: Yes — probably extend the message noting that JS functions cannot be Java-ified, so it is clear in what context the error is thrown.

florian-h05:
So wrt to the duplicate code for handling the conversion - I don't see a problem in that, it is not that many lines and very simple code.

I tried this out and it looks good. Will provide a jsify function as a draft PR soon ...

florian-h05:
Can you please move runAsync to a dedicated PR?

Nadahar:
> Can you please move runAsync to a dedicated PR?

I can't move it to a dedicated PR, since it uses `javaify()`, but I've removed the commit.

Nadahar:
> So wrt to the duplicate code for handling the conversion - I don't see a problem in that, it is not that many lines and very simple code.

There is no duplicate code actually, because this is recursive - so using the existing functions wouldn't have worked. It calls itself while iterating.

src/utils.js
```diff
@@ -144,6 +144,14 @@ function javaify (val) {
     return val;
   }
 
+  if (typeof val !== 'object') {
+    return val;
+  }
```
Nadahar: You just disabled the function check.... now it will return a function as it is.

Nadahar: I've pushed a new commit that reorder them.

Nadahar:
I don't know how long you've come, but I've tried to "reverse" `javaify` to `jsify`, and come up with this. Not sure if everything works or is correct, but it might be a starting point:

```js
/**
 * Recursively convert any Java type to native JavaScript types when applicable. What cannot be converted remains Java objects.
 *
 * @param {*} val the Java object to convert
 * @returns {*} The converted value using native JS-types where possible.
 */
function jsify (val) {
  if (val === null || val === undefined) {
    return null;
  }

  if (!Java.isJavaObject(val) || typeof val !== 'object') {
    return val;
  }

  if (val instanceof Java.type('java.lang.Number')) {
    return val.doubleValue();
  }

  // Convert java.time objects
  const className = val.getClass().getName();
  if (className === 'java.time.LocalDate') {
    return LocalDate.parse(val.toString());
  }
  if (className === 'java.time.ZonedDateTime') {
    return ZonedDateTime.parse(val.toString());
  }
  if (className === 'java.time.LocalDateTime') {
    return LocalDateTime.parse(val.toString());
  }
  if (className === 'java.time.Instant') {
    return Instant.ofEpochMilli(val.toEpochMilli());
  }
  if (className === 'java.time.Duration') {
    return Duration.ofNanos(val.toNanos());
  }

  // Convert Lists and arrays
  if (val instanceof Java.type('java.util.List') || val.getClass().isArray()) {
    const jsArray = [];
    for (const item of val) {
      jsArray.push(jsify(item));
    }
    return jsArray;
  }

  // Convert Sets
  if (val instanceof Java.type('java.util.Set')) {
    const jsSet = new Set();
    const iterator = val.iterator();
    while (iterator.hasNext()) {
      jsSet.add(jsify(iterator.next()));
    }
    return jsSet;
  }

  // Java Maps to objects
  if (val instanceof Java.type('java.util.Map')) {
    const jsObj = {};
    const iterator = val.entrySet().iterator();
    while (iterator.hasNext()) {
      const entry = iterator.next();
      jsObj[String(entry.getKey())] = jsify(entry.getValue());
    }
    return jsObj;
  }

  // Return everything else as Java objects
  return val;
}
```

florian-h05:
It‘s pretty similar to what I have. I am bit hesitant how to handle java.time.* types as utils.js shoudn‘t become dependent on time.js …

src/utils.js
```diff
@@ -127,6 +129,92 @@ function javaSetToJsSet (set) {
...
+    const javaSet = new LinkedHashSet();
+    val.forEach((value) => {
+      javaSet.add(javaify(value));
+    });
+    return javaSet;
+  }
+
+  // Convert JS objects
+  const map = new LinkedHashMap();
+  for (const key in val) {
```
florian-h05: If you use Object.keys() (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/keys) you don’t need the hasOwnProperty check.

florian-h05 (commented):
One comment, apart from this LGMT!

florian-h05 (approved):

Nadahar:
> I am bit hesitant how to handle java.time.* types as utils.js shoudn‘t become dependent on time.js …

I have nothing to contribute is such deliberations... but in my test, I have this on top:

```js
const ZonedDateTime = require('@js-joda/core').ZonedDateTime;
const LocalDate = require('@js-joda/core').LocalDate;
const LocalDateTime = require('@js-joda/core').LocalDateTime;
const Instant = require('@js-joda/core').Instant;
const Duration = require('@js-joda/core').Duration;
```

`time.js` isn't mentioned anywhere in the file.

PR #547, opened 2026-07-14
Add `types:bundle` script to bundle type definitions into a single file

florian-h05:


florian-h05:
@digitaldan We need to figure out a way to make the openhab.d.ts file accesible to the MCP server to supercharge code generation of openHAB rules.

digitaldan:
Hmm, thats an interesting challenge..... I would need to think on how we could do that in a non hacky way,  unless you have some ideas there.  If there was a standard cache path, un-versioned, that the MCP could read from, that might work....

florian-h05:
What if we hosted it somewhere on GitHub or our website, with a redirect from a static path on our website to the actual source? This way we could control which version of the type definition the MCP add-on gets without modifying the MCP code. 
I am thinking of something comparable to how I handle docs versioning when loading add-on docs and developer sidebar help docs in Main UI:

https://github.com/openhab/openhab-website/blob/15ea0914e3bbf86ac38ffae95433d5409467c366/.vuepress/_redirects#L32

Actually it would be great to have the type definitions somewhere where the docs could link to them — when we could advise users to provide the type defs when interacting with ChatGPT, Gemini etc in the browser.

florian-h05:
@digitaldan You can now get the type defs from this URL:

```
https://www.openhab.org/link/docs-src/5.3.x/addons/automation/jsscripting/res/openhab.d.ts
```

This provides versioning support, it always points to the openhab.d.ts for the openhab-js version shipped with openHAB 5.3.x.

PR #549, opened 2026-07-16
[cache] `javaify()` & `jsify()` JS objects in shared cache & [rules] Make `EventObject::raw` contain raw data

Nadahar:
This isn't meant to be used as is - I just found it easier to show what I'm thinking that trying to explain it.

I've dealt with both `_getTriggeredData()` and the shared cache, but I am somewhat confused about the `raw` property as I thought that was supposed to be "unjsified".


src/rules/rules.js
```diff
@@ -495,7 +495,7 @@ function _collapseInputMap (input) {
  * @returns {EventObject}
  */
 function _getTriggeredData (rawInput, javaEventBackwardCompat = false) {
-  const inputAsJsObj = javaMapToJsObj(rawInput);
+  const inputAsJsObj = jsify(rawInput);
```
florian-h05: I don't think there is a need to jsify this here - it is only used internally in this method of the helper library and might possibly brake the currently working code.

Nadahar:
@florian-h05 Please take a look at what I've suggested here as well. It's primarily about the shared cache, but it also modifies `getTriggeredData()` a bit, in that it calls `jsify()` on the data to be jsified instead of `javaMapToJsObj()`, which I assume only converts the outer map. Then I also put the raw map directly in `raw`, but perhaps it should be treated with `javaMapToJsObj()` first?

src/cache.js
```diff
@@ -48,12 +53,11 @@ class JSCache {
    * @param {*} value value to be associated with the specified key
    * @returns {*|null} the previous value associated with the key, or null if there was no mapping for key
    */
-  put (key, value) {
+  put (key, value, jsifyResult = true) {
+    const isShared = this.#isSharedCache();
     // see https://www.graalvm.org/latest/reference-manual/js/JavaScriptCompatibility/ for Java. docs
```
florian-h05: This comment can be removed.

src/cache.js
florian-h05, line 1: Can you please update the methods' JSDoc comments for the new param?

You can define an optional param with a default like this:

`@param {type} [name=default] description`

florian-h05:
> Then I also put the raw map directly in raw, but perhaps it should be treated with javaMapToJsObj() first?

I would actually keep `raw` the raw Java Map according to its name. Not sure when or why I changed it to a JS object ...
WDYT?

src/rules/rules.js
Nadahar: OK - my intention was to jsify the content that ended up in the hands of the user - but was outside `raw`. My idea was that since it's recursive, it will convert objects in the "whole tree", so whatever you do with it later, the objects will already be JS.

src/cache.js
Nadahar: Sure, this was a suggestion, so I didn't modify JSDocs util I know that the idea is "accepted".

Nadahar:
> I would actually keep `raw` the raw Java Map according to its name. Not sure when or why I changed it to a JS object ...
> WDYT?

Yes, I agree, my expectation would be that it should be "raw". I also think it makes little difference in most cases, since Graal already lets you do a lot of things "natively" with a Java map.

src/cache.js
Nadahar: Done.

src/cache.js
Nadahar: Done.

src/rules/rules.js
```diff
@@ -496,7 +496,7 @@ function _collapseInputMap (input) {
  */
 function _getTriggeredData (rawInput, javaEventBackwardCompat = false) {
   const inputAsJsObj = javaMapToJsObj(rawInput);
-  const input = _collapseInputMap(inputAsJsObj);
+  const input = jsify(_collapseInputMap(inputAsJsObj));
```
Nadahar: @florian-h05 What about this? Or, do you simply think that it's "risky" to jsify the content of `input`?

florian-h05: I don't see much benefit in doing so ... this method outputs a pure JS object (except for the `raw` property), jsifying here could only break the method internally and adds overhead.

src/rules/rules.js
florian-h05: Hmm, might allow us to rewrite the function body a bit, but TBH my motivation to test that change is not that high ...

src/rules/rules.js
Nadahar: I don't understand how. The objects inside `input` isn't converted to JS objects, are they? I thought that only applied to the "event object" itself, not all the other content that comes from other modules.

src/rules/rules.js
Nadahar: Perhaps @rkoshak is interested? We can just drop the whole thing for me (in `getTriggeredData()`), I just thought that since we now have `jsify()`, we might as well use it to make the context objects even easier to work with.

florian-h05: The event object is already 100% JS. `ctx` is something only existing in ScriptActionHandler, nothing that exits in `JSRule` or rule builder and nothing that we can handle from this place here.
When you call a `JSRule` with `rules.runRule`, the passed in context will be in the `.raw` attribute of the event object. I think we can expect JSRule users to be able to call `jsify` themselves if needed.

src/rules/rules.js
florian-h05: The objects inside `input` are allowed to remain Java - the code down below extracts data from them and provides that data as JS types on the event object.

Nadahar: Ok, so what you're saying is that you want to keep it that way.

That's fine too, users can call `jsify()` whenever that suits them.

Nadahar: I've removed it.

Nadahar:
@florian-h05 It sounds like you think the rest of this is good, so I took it out of draft. That leaves only the change to `input.raw`, otherwise the changes only apply to the shared cache.

Nadahar:
Also, if somebody you check/test/verify https://github.com/openhab/openhab-webui/pull/4315 and it's OK, I could give creating blocks for `javaify()` and `jsify()` a shot.

src/cache.js
```diff
@@ -29,41 +30,46 @@ class JSCache {
    *
    * @param {string} key the key whose associated value is to be returned
    * @param {function} [defaultSupplier] if the specified key is not already associated with a value, this function will return a default value
+   * @param {boolean} [jsifyResult=true] whether or not the result will be automatically converted to JavaScript objects when possible.
```
florian-h05:
```suggestion
   * @param {boolean} [jsifyResult=true] whether or not the result will be automatically converted to JavaScript objects when possible
```

src/cache.js
```diff
@@ -29,41 +30,46 @@ class JSCache {
...
     if (typeof defaultSupplier !== 'function') return null;
     const supplied = defaultSupplier();
     if (supplied === null) return null; // do not store null values in cache
-    this.#valueCache.put(key, supplied);
-    return supplied;
+    this.#valueCache.put(key, isShared ? javaify(supplied) : supplied);
+    return isShared && jsifyResult ? jsify(supplied) : supplied;
   }
 
   /**
```
florian-h05, line 50: I think we should add to the JSDoc that JS types are automatically javaified when put into the shared cache as this is technically required.

src/cache.js
```diff
@@ -29,41 +30,46 @@ class JSCache {
...
    * Associates the specified value with the specified key.
    *
    * @param {string} key key with which the specified value is to be associated
    * @param {*} value value to be associated with the specified key
+   * @param {boolean} [jsifyResult=true] whether or not the result will be automatically converted to JavaScript objects when possible.
```
florian-h05:
```suggestion
   * @param {boolean} [jsifyResult=true] whether or not the result will be automatically converted to JavaScript objects when possible
```

src/cache.js
Nadahar: Please review what I wrote, wasn't quite sure how to put it.

florian-h05 (approved):
LGTM, thanks!

PR #560, opened 2026-08-13
[cache] Make javaify() optional when putting to shared cache & [rules] Add collapsed map to EventObject.inputs

### LINKED ISSUE #559: Functions cannot be put into shared cache anymore
Hi,

I have a setup where I put functions into the shared cache so they can be used as a callback by other rules. This is used to dynamically execute actions using Pushover messages.
It worked fine, but with updating to OH 5.2.1 and the latest openhab-js this fails. This is most likely caused by the newly introduced `javaify` method. It would be great if the old behavior would still be possible.

Nadahar:
There have been some discussions both on the forum and here: https://github.com/rkoshak/openhab-rules-tools/issues/12 about the implicit `javaify()` when putting into the shared cache. I didn't think there was a need for it to be optional, because I didn't think there existed cases where it would work to put JS objects in the shared cache.

It turns out that was wrong. When a rule wants to share something between a condition and action, it can't use the private cache, but can safely put JS objects there since the rule is handled by the same thread. There is also the situation where multiple rules are defined in the same JS file, where it might be safe to share JS objects among those rules.

So, I've added an optional parameter that defaults to true for `javaify()` when putting to the cache.

In addition, I've added the collapsed map to `EventObject.inputs`, so that unprefixed entries are still easily available. Think of it as a suggestion from my side, but I think it makes sense. I thought it was already available, and was surprised when I found that the collapsed map we built was never forwarded to the user.

Fixes #559.


Owlbertz:
I guess that would solve #559.

src/cache.js
```diff
@@ -51,21 +51,22 @@ class JSCache {
    * Associates the specified value with the specified key.
    *
    * @param {string} key key with which the specified value is to be associated
-   * @param {*} value value to be associated with the specified key. The value will be run through javaify() if the cache is shared, since the shared cache can't contain JavaScript objects.
-   * @param {boolean} [jsifyResult=true] whether or not the result will be automatically converted to JavaScript objects when possible
+   * @param {*} value value to be associated with the specified key.
+   * @param {boolean} [jsifyResult=true] whether or not the result will be automatically converted to JavaScript objects when possible, if the cache is shared
```
florian-h05, line 57: Can you please swap the param doc order, so it matches the actual function signature?

src/cache.js
```diff
@@ -51,21 +51,22 @@ class JSCache {
...
+   * @param {boolean} [javaify=true] whether or not the value will be automatically converted to Java objects if the cache is shared
    * @returns {*|null} the previous value associated with the key, or null if there was no mapping for key
    */
-  put (key, value, jsifyResult = true) {
+  put (key, value, javaify = true, jsifyResult = true) {
     const isShared = this.#isSharedCache();
-    const result = this.#valueCache.put(key, isShared ? javaify(value) : value);
+    const result = this.#valueCache.put(key, isShared && javaify ? javaify(value) : value);
```
florian-h05: Doesn't javaify param interfere with javaify method?

src/rules/rules.js
```diff
@@ -496,32 +496,33 @@ function _collapseInputMap (input) {
...
-  const event = input.event;
+  const event = inputs.event;
   /**
    * @type {EventObject}
    */
   const data = {};
 
   // Add input to data to pass through any properties not captured below
   data.raw = rawInput;
+  data.inputs = inputs;
```
florian-h05, line 510: Can you please add this to the `EventObject` typedef (JSDoc) on the top of this file:

You can type inputs as `Record<string, unknown>`

src/cache.js
Nadahar: I haven't tested the code, not sure exactly how to, but "from where I come" `()` isn't optional, so there's no potential for conflict. That might not be true here, so I'll rename the option to `javaifyValue`.

src/rules/rules.js
Nadahar: I'm very unsure about this, `raw` doesn't use `Record`. Am I on the right track here?

```js
 * @property {*} raw original contents of the event including data passed from a calling rule
 * @property {Record<string, unknown>} inputs the same as 'raw' except that keys exist without prefix, making them easier to find
 * @property {*} payload if provided by event: payload of event in Java data types
```

Nadahar: @florian-h05 I don't know if you never saw the comment, but I'm holding off my commit until I know if this is correct or not.

Nadahar: @florian-h05 Is there some bug with GitHub again? I'd think it would be easy to answer if what I suggested above is OK or not...

src/cache.js
florian-h05: It indeed interferes with the method.
Method params can also be functions, and as JS doesn't know types it isn't able to differentiate between the javaify method and the javaify param. The javaify param "shaded" the method, calling `javaify()` then assumes the param is a function and can be executed.

src/rules/rules.js
florian-h05: `raw` is the raw Java map, we actually have a type defined for that so you can use this: `JavaMap`
`inputs` is a JS object with "collapsed" keys, so it can be types as a `Record<string, unknown>`.
Note that your JSDoc isn't entirely correct about inputs, in addition to collapsing keys it is a JS obj instead of a Java map.

And sorry for the late answer, was away for several days.

Nadahar: You kind of ended up confusing me even more, this is what I came up with, but I'm not at all sure that I understood that `JavaMap` part right:

```js
 * @property {JavaMap} raw original contents of the event including data passed from a calling rule
 * @property {Record<string, unknown>} inputs the same as 'raw' except that it's a JS object instead of a Java Map, and the keys exist without prefix
 * @property {*} payload if provided by event: payload of event in Java data types
```

florian-h05: It looks good to me that way.

florian-h05 (approved):
LGTM, thanks!

Can you please split the changes to cache and rules into two dedicated commits for rebase merging? I don't want to squash them.

Nadahar:
> Can you please split the changes to cache and rules into two dedicated commits for rebase merging? I don't want to squash them.

Done

PR #566, opened 2026-09-07
Convert numeric states to string when persisting

### LINKED ISSUE #565: ItemPersistence.persist() fails for numeric states
## The problem

`ItemPersistence.persist()` cannot store a numeric state. Persisting a `TimeSeries` whose states are JS numbers throws:

```
TypeError: invokeMember (parseState) on org.openhab.core.types.TypeParser failed due to:
Cannot convert '38.16'(language: Java, type: java.lang.Double) to Java type 'java.lang.String':
Invalid or lossy primitive coercion.
    at persist (item-persistence.js:203)
```

Numbers are explicitly documented as accepted — the JSDoc for `persist` declares `state` as `{string|number|time.ZonedDateTime|Quantity|HostState}`, and `TimeSeries.add` likewise — so this is a contract mismatch rather than misuse.

## Cause

`_toOpenhabPrimitiveType` returns numbers unchanged ([`src/helpers.js#L41-L43`](https://github.com/openhab/openhab-js/blob/main/src/helpers.js#L41-L43)):

```js
if (typeof value === 'number' || typeof value === 'string') {
  return value;
}
```

`#persistTimeSeries` then passes that straight into `TypeParser.parseState` ([`src/items/item-persistence.js#L285`](https://github.com/openhab/openhab-js/blob/main/src/items/item-persistence.js#L285)):

```js
ts.add(timestamp, TypeParser.parseState(acceptedDataTypes, _toOpenhabPrimitiveType(state)));
```

`TypeParser.parseState(List<Class<? extends State>>, String)` takes a Java `String`, and GraalJS will not coerce a `java.lang.Double` into one.

`#persistGivenState` (lines 267 and 270) looks affected the same way: `PersistenceExtensions.persist` offers only `(Item, ZonedDateTime, State)` and `(Item, ZonedDateTime, String)` overloads, and a JS number matches neither. I worked around that path pre-emptively in my own code, so I have not observed it failing in production — only the `TimeSeries` path.

## Steps to reproduce

```js
const ts = new items.TimeSeries('REPLACE');
ts.add(time.ZonedDateTime.now().plusHours(1), 38.16);   // a number, as documented
items.getItem('SomeNumberItem').persistence.persist(ts, 'influxdb');
```

Passing `'38.16'` instead works — `parseState` parses the text back into a `DecimalType`.

## Suggested fix

Stringify at the persistence call sites rather than in `_toOpenhabPrimitiveType`, which `sendCommand` and `postUpdate` also use and where numbers are fine:

```js
TypeParser.parseState(acceptedDataTypes, String(_toOpenhabPrimitiveType(state)))
```

I have a PR ready with this fix plus unit tests, including mocks for `PersistenceExtensions`, `TimeSeries` and `TypeParser` (there is currently no `test/items/item-persistence.spec.js`). Will open it right after this.

## Your environment

- openHAB 5.0.2, openhab-js 5.16.2 — code unchanged on `main` as of 5.21.0
- Found while persisting day-ahead electricity prices as a forecast `TimeSeries` with future timestamps

lkleen:
Fixes #565

## What this fixes

`ItemPersistence.persist()` could not store a numeric state, although the JSDoc for `persist()` and for `TimeSeries.add()` both document `number` as accepted:

```
TypeError: Cannot convert '38.16'(language: Java, type: java.lang.Double)
to Java type 'java.lang.String': Invalid or lossy primitive coercion.
```

`_toOpenhabPrimitiveType` returns numbers unchanged, and `#persistTimeSeries` passes the result to `TypeParser.parseState(List<Class<? extends State>>, String)`, whose second parameter is a Java `String`. GraalJS declines the conversion. `#persistGivenState` is affected the same way — `PersistenceExtensions` offers only `(Item, ZonedDateTime, State)` and `(Item, ZonedDateTime, String)` overloads, and a number matches neither.

## Approach

The stringification is done in a new module-local `_toOpenhabStringType`, applied at the three persistence call sites, rather than inside `_toOpenhabPrimitiveType`.

That placement is deliberate: `sendCommand` and `postUpdate` also use `_toOpenhabPrimitiveType` and pass its result to APIs that *do* accept numbers, so changing the shared helper would be a wider behavioural change than this bug warrants.

`String()` is a no-op for every other branch of `_toOpenhabPrimitiveType` — the `'NULL'` / `'UNDEF'` sentinels and the `toOpenHabString()` / `toString()` results are already strings — so only the numeric path changes.

## Tests

`test/items/item-persistence.spec.js` is new; there was no spec for this module. It needed mocks for `PersistenceExtensions`, `TimeSeries` and `TypeParser`, which are added to `test/openhab.mock.js` and registered in `test/jest.setup.js`.

The `TypeParser.parseState` mock **throws on a non-string second argument**, mirroring GraalJS host interop. That is what makes this a real regression test: with the fix reverted, three of the five new tests fail, one of them with the exact error from the issue, rather than the bug being invisible until it reaches openHAB.

Five tests cover numeric states via `TimeSeries` and via `persist(timestamp, state[, serviceId])`, plus the unchanged string and `null` / `undefined` paths.

`npm run lint` is clean and the full suite passes at 200 tests, up from 195.


lkleen:
This fix is AI generated, but I actually stumbled into this issue while I was working on my openHAB setup. As the diagnosis sounds reasonable to me I decided to tell the AI to generate a fix for it and open a PR. The generated code also look reasonable to me but I'm not a regular openHAB developer so I can't pre-review with the conventions in mind I need to apply to proably

